PR #26, opened 2022-04-25
feat(Storage): Smoother loading state

ariser:
## Suggested change
Feature: Storage page

Current state:
- The table on the Storage page does not indicate its loading state, this is confusing for large tables with long loading times.
- Filter controls over the table are hidden while the table is loading, which does not make sense since they don’t depend on the table data

Suggested change:
- Always show the filter controls, and only put the table in loading state
- Replace Loader with Skeleton to improve visuals
- Avoid requests races by using `cuncurrentId` for requests for table data

## Bug fix
This PR also fixes a bug found while fine-tuning the requests.

What was wrong:
In rare cases there could be ghost background requests.
The Storage page uses AutoFetcher to refresh table data in the background.
The issue occurs in case there are consequent calls of `.stop()` and then `.start()`, while the background request is in progress. The method `.fetch()` only checks if autofetcher is active before initiating the request, but not afterwards. Thus it continue to fetch an old request, which had been stopped, along with a new one.

How is it fixed:
AutoFethcer now tracks the number of times it was started. After the request is completed, a check is performed, whether the current launch has the same sequential number as the latest one in AutoFetcher instance. If not, the next fetch is not initiated.

### DIFF
--- a/src/containers/Storage/Storage.js
+++ b/src/containers/Storage/Storage.js
@@ -3,10 +3,11 @@ import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import cn from 'bem-cn-lite';
 import DataTable from '@yandex-cloud/react-data-table';
-import {Loader, RadioButton, Label} from '@yandex-cloud/uikit';
+import {RadioButton, Label} from '@yandex-cloud/uikit';
 
 import StorageFilter from './StorageFilter/StorageFilter';
 import {AutoFetcher} from '../../utils/autofetcher';
+import {TableSkeleton} from '../../components/TableSkeleton/TableSkeleton';
 
 import {
     getStorageInfo,
@@ -68,17 +69,14 @@ class Storage extends React.Component {
             storageType,
             setHeader,
             getNodesList,
-            getStorageInfo,
         } = this.props;
 
         this.autofetcher = new AutoFetcher();
         getNodesList();
         if (tenant || nodeId) {
             setVisibleEntities(VisibleEntities.All);
-            getStorageInfo({
-                tenant,
+            this.getStorageInfo({
                 filter: FILTER_OPTIONS.All,
-                nodeId,
                 type: storageType,
             });
         } else {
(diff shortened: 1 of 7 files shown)


src/containers/Storage/Storage.js
```diff
@@ -80,6 +79,8 @@ class Storage extends React.Component {
                 filter: FILTER_OPTIONS.All,
                 nodeId,
                 type: storageType,
+            }, {
+                concurrentId: 'getStorageInfo',
```
Raubzeug, line 83: Let's make this method as a separate function to avoid code duplication.

src/containers/Storage/Storage.js
```diff
@@ -159,8 +170,20 @@ class Storage extends React.Component {
 
     renderLoader() {
         return (
-            <div className={b('loader')}>
-                <Loader size="m" />
+            <div className={b('skeleton')}>
```
Raubzeug, line 173: We have a lot of tables in this apps that may need a Skeleton in the near future. I suggest to make Table Skeleton as a separate component.

src/store/reducers/storage.js
```diff
@@ -22,14 +22,15 @@ export const StorageTypes = {
...
+// works exactly the same as FETCH_STORAGE, but does not trigger loading state
+const BG_FETCH_STORAGE = createRequestActionTypes('storage', 'BG_FETCH_STORAGE');
 const SET_INITIAL = 'storage/SET_INITIAL';
 const SET_FILTER = 'storage/SET_FILTER';
 const SET_VISIBLE_GROUPS = 'storage/SET_VISIBLE_GROUPS';
 const SET_STORAGE_TYPE = 'storage/SET_STORAGE_TYPE';
 
 const initialState = {
     loading: true,
-    wasLoaded: false,
```
Raubzeug, line 32: I'm not sure it's good idea because in all other places of the app background refresh is implemented with wasLoaded.
Why don't you like this approach?

Raubzeug (changes requested):

src/store/reducers/storage.js
ariser: It just doesn’t apply. `wasLoaded` indicates what the name implies: whether data was loaded for the first time or not. This is not enough information to handle the case in which app should behave differently for manual requests and for background automatic requests.

Moreover, if this approach is used in other places in the app, we should redo those places :)

src/containers/Storage/Storage.js
ariser: what method are you talking about?

src/store/reducers/storage.js
ariser: Discussed in PM, the issue is specific to this module, in other places `wasLoaded` resets every time data is refetched manually. Will do the same here.

Raubzeug (dismissed):

Raubzeug (approved):

PR #30, opened 2022-05-04
fix: reducer clusterInfo should not be used

Raubzeug:
Reducer clusterInfo is redundant and should be removed.


src/containers/Header/Header.tsx
```diff
@@ -33,7 +33,7 @@ function Header() {
         useSelector((state: any) => state);
 
     const clusterName: string = useSelector(
-        (state: any) => state.cluster.data?.Name || state.clusterInfo.title,
+        (state: any) => state.cluster.data?.Name || state.clusterInfo?.title,
```
ariser, line 36: do we really want to keep the usage of a reducer that isn't in the code base anymore?

ariser (changes requested):

Raubzeug:
Yes, now it is needed in internal version.

src/containers/Header/Header.tsx
Raubzeug: Yes, now it is needed in internal version.

ariser: Keeping in the open source version something that is only required in the internal version doesn't seem right to me. This selector should accept some parameter, and internal version should use it to fallback to `clusterInfo.title`. And in open source we should only keep `state.cluster`.

I can approve this PR if we agree to change this selector in the future

Raubzeug: I totally agree with you. It seems not very difficult to achieve this, so I fixed the issue.

ariser (approved):

PR #32, opened 2022-05-04
Scroll-related layout fixes

ariser:
Fixes introduced in this PR:
- sticky header in the ACL area of Tenant feature scrolled up after some threshold; now it is always stuck to the top side
- storage table content was visible next to the filters row for tables with horizontal scroll, after scrolling to the right; now it is not
- scrollbars on the Tenant page had messy behavior; made it more consistent
- added missing right padding on the Node storage page

### DIFF
--- a/src/components/NodesViewer/NodesViewer.scss
+++ b/src/components/NodesViewer/NodesViewer.scss
@@ -5,11 +5,7 @@
     @include flex-container();
 
     &__controls {
-        display: flex;
-        align-items: center;
-        gap: 12px;
-
-        margin: 0 0 8px;
+        @include controls;
     }
 
     &__name {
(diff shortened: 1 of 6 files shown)


Raubzeug (changes requested):
With these changes applied only Strorage tab in Diagnostics looks good. All other tabs content had lost gap between filters and content.

Raubzeug:
There's one trickery: Nodes and Storage components are used in different places: a part of Diagnostics page in Tenant and an independent page (from main navigation). So layout should satisfy both cases.

ariser:
I tried to review all the use cases. Will double check it again.

Raubzeug (changes requested):
Now on Storage page horizontal scroll appears right under the storage table. It looks not good, better do it as on Nodes page - in the bottom of the whole page.

Raubzeug (approved):

Raubzeug (approved):

PR #37, opened 2022-05-05
fix(Diagnostics): layout

Raubzeug:
In this PR I want to fix problems below:
Problem 1: when we view Diagnostic-details modal (in Database-Diagnostics-Overview-Healthcheck when it is something wrong with database), its height varies depends on content height, and it looks not good when modal's height jumps up and down.
Problem 2: in Database-Diagnostics-Overview-Healthcheck page title, badge and update-button should be centred.

### DIFF
--- a/src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.scss
+++ b/src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.scss
@@ -37,23 +37,22 @@
 
     &__status-wrapper {
         display: flex;
-        gap: 8px;
+        align-items: baseline;
 
         margin-bottom: 15px;
+        gap: 8px;
     }
 
     &__preview-title {
-        display: flex;
-        align-items: center;
-
         font-weight: 600;
+        line-height: 24px;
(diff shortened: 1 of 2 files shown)


src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.scss
```diff
@@ -35,10 +43,16 @@
...
     &__preview-title {
+        display: flex;
+        align-items: center;
+
         font-weight: 600;
     }
 
     &__self-check-status-indicator {
+        display: flex;
+        align-items: center;
```
ariser, line 54: no reason for these elements to be flex, and the alignment should be baseline, not center
i'd rather 
```scss
.healthcheck {
    &__status-wrapper {
        ...
        align-items: baseline;
    }
}
```

ariser (changes requested):

ariser (changes requested):

ariser (dismissed):

ariser (approved):

PR #43, opened 2022-05-12
feat(Tenant): save initial tab preference

ariser:
Save the current tab after switching to it to open it initially next time


src/containers/Tenant/ObjectGeneral/ObjectGeneral.tsx
```diff
@@ -63,6 +68,7 @@ function ObjectGeneral(props: ObjectGeneralProps) {
                         );
                     }}
                     allowNotSelected
+                    onSelectTab={(id) => localStorage.setItem(TENANT_GENERAL_TABS_LOCAL_STORAGE, id)}
```
Raubzeug, line 71: I think it would be better to use "setSettingValue" and "getSettingValue" reducers - https://github.com/ydb-platform/ydb-embedded-ui/blob/b213b97a6af5c23f56294a00c33c04d91420b8a2/src/store/reducers/settings.js#L53-L67 (user settings in state should be tuned too - https://github.com/ydb-platform/ydb-embedded-ui/blob/b213b97a6af5c23f56294a00c33c04d91420b8a2/src/store/reducers/settings.js#L17

Raubzeug (changes requested):

Raubzeug (approved):

PR #50, opened 2022-05-19
Use ydb-ui-components/NavigationTree for tenants schema

ariser:
This replaces custom components for displaying tree with `NavigationTree` from `ydb-ui-components`.
Along with this, a followup refactoring and styles adjustment.


package.json
```diff
@@ -105,5 +106,8 @@
     "react-dom": "^17.0.2",
     "stylelint": "^14.3.0",
     "typescript": "^4.5.5"
+  },
+  "peerDependencies": {
+    "@yandex-cloud/browserslist-config": "^1.0.1"
```
ariser, line 110: Questionable.
I added `@yandex-cloud/browserslist-config` in `ydb-ui-components` peerDependencies because browserslist in package.json of `ydb-ui-components` extends a config from `@yandex-cloud/browserslist-config`. This causes build of `ydb-embedded-ui` to crash because of a missing dependency. Adding the package to peer deps doesn't solve the problem, but provides a hint to install this package.

But I assume that build of any project that depends on `ydb-embedded-ui` will similarly fail for the same reason. And I don't like the solution with specifying this peer dependency. It's kinda unhelpful.

I guess we need to solve the issue with crashing build some other way.

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -195,13 +201,36 @@ function ObjectSummary(props: ObjectSummaryProps) {
...
                 <div className={b('tree-header')}>
                     <div className={b('tree-title')}>Navigation</div>
                 </div>
                 <div className={b('tree')}>
                     {tenantData && (
-                        <SchemaNode fullPath={tenantName as string} data={tenantData} isRoot />
+                        <NavigationTree
+                            rootState={{
+                                path: tenantName as string,
+                                name: tenantName as string,
```
ariser, line 213: A question: is tenant name always also tenant path? Or is it better to read the path from the `Path` field?

src/containers/Tenant/utils/schemaActions.ts
```diff
@@ -0,0 +1,119 @@
+import qs from 'qs';
```
ariser, line 1: This file is based on `SchemaNodeActions`, not identical

src/containers/Tenant/utils/schema.ts
```diff
@@ -0,0 +1,17 @@
+import type {NavigationTreeNodeType} from "ydb-ui-components";
+
+const DB_TYPES = new Set(['EPathTypeSubDomain']);
+const TABLE_TYPES = new Set(['EPathTypeTable', 'EPathTypeOlapTable']);
+const DIR_TYPES = new Set(['EPathTypeDir', 'EPathTypeOlapStore']);
```
ariser, line 5: needs double checking

src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return (
+        <NavigationTree
+            rootState={{
+                path: rootPath,
+                name: rootName,
+                type: calcNavigationTreeType(rootType),
+                collapsed: false,
+            }}
+            fetchPath={fetchPath}
+            getActions={getActions(dispatch, history, handleActivePathUpdate)}
```
ariser, line 55: I don't really like this `getActions` function signature
I already did some refactoring, but there is still place for improvement
if any idea comes to mind, please share

Raubzeug:
При первом заходе на базу дерево подгружается некорректно.

Raubzeug:
When user closes table preview, if there is no query "under" it, should minimise empty split panel.

Raubzeug (changes requested):

ariser:
> When user closes table preview, if there is no query "under" it, should minimise empty split panel.

Decided in PM to make this a separate change, this is a new feature

> The first time we enter the database, the tree loads incorrectly.

uploaded a fix for this

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -7,26 +7,20 @@ import qs from 'qs';
 import _ from 'lodash';
 
 import {Button, HelpTooltip, Loader, Tabs} from '@yandex-cloud/uikit';
-//@ts-ignore
```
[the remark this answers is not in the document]
ariser: This project properly builds. There may be issues with the internal build, will deal with it there, or will revert these ignores in case it can't be dealt with

Raubzeug (dismissed):

Raubzeug (approved):

PR #51, opened 2022-05-20
Improve Save Query UX

ariser:
This introduces 4 updates:
1. Query deletion can now be confirmed by hitting Enter
2. Query saving can now be confirmed by hitting Enter
3. Query name field is now initially focused, allowing immediate typing
4. The warning for losing saved queries upon cookies deletion is now replaced with a note about how queries are saved in the save dialog

More about 4:
The reasons why it needs to be changed:
- the message is incorrect: queries are saved not in cookies, but in local storage
- the message under the question mark icon requires an extra action to be seen; the message in the save dialog does not

### DIFF
--- a/src/containers/Tenant/QueryEditor/SaveQuery/SaveQuery.scss
+++ b/src/containers/Tenant/QueryEditor/SaveQuery/SaveQuery.scss
@@ -1,17 +1,19 @@
-@import '../../../../styles/mixins.scss';
-
 .kv-save-query {
-    &__dialog-body {
+    &__dialog-row {
         display: flex;
         align-items: flex-start;
+
+        & + & {
+            margin-top: var(--yc-text-body-line-height);
+        }
     }
     &__field-title {
         margin-right: 12px;
 
         font-weight: 500;
-        line-height: 32px;
+        line-height: 28px;
         white-space: nowrap;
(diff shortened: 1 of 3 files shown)


src/containers/Tenant/QueryEditor/SaveQuery/SaveQuery.scss
```diff
@@ -1,17 +1,19 @@
...
+        & + & {
+            margin-top: var(--yc-text-body-line-height);
+        }
     }
     &__field-title {
         margin-right: 12px;
 
         font-weight: 500;
-        line-height: 32px;
+        line-height: 28px;
```
ariser, line 14: the height of the text field
this artificially aligns the label with the field by baseline
align-items: baseline for the parent doesn’t work without extra refactoring

src/containers/Tenant/QueryEditor/SaveQuery/SaveQuery.scss
```diff
@@ -1,17 +1,19 @@
...
         white-space: nowrap;
-        @include body2-typography();
```
ariser, line 14: i consider the label doesn’t require a bigger font here, especially with another text line next to it

Raubzeug (approved):

PR #59, opened 2022-05-25
fIx Tenant tabs navigation

ariser:
there are two commits:

### URL params replacement fix
redux-location-state tends to overwrite changes made with HistoryAPI, if push to history is made simultaneously with modifying URL with redux-location-state. Better not to mix these two approaches.

### Schema navigation fix
In dev environment, 2nd level tabs on Diagnostics sometimes don't update after changing the current node in the schema. Investigated and fixed.

### DIFF
--- a/src/containers/Tenant/Diagnostics/Diagnostics.tsx
+++ b/src/containers/Tenant/Diagnostics/Diagnostics.tsx
@@ -79,7 +79,7 @@ function Diagnostics(props: DiagnosticsProps) {
             pages = TABLE_PAGES;
         }
         return pages;
-    }, [props.type]);
+    }, [props.type, isDatabase]);
 
     const forwardToDiagnosticTab = (tab: GeneralPagesIds) => {
         dispatch(setDiagnosticsTab(tab));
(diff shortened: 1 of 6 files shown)


src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -76,15 +78,10 @@ function Diagnostics(props: DiagnosticsProps) {
             pages = TABLE_PAGES;
         }
         return pages;
-    }, [props.type]);
+    }, [props.type, isDatabase]);
```
ariser, line 81: I suppose this fix is primarily for dev environment, where the root is considered to be a database, though having the type of a directory. But with the present logic this change is only reasonable, since it assumes this kind of situation, when `isDatabase` conflicts with `props.type`.

src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -78,7 +78,7 @@ function Diagnostics(props: DiagnosticsProps) {
...
```
ariser, line 82: I suppose this fix is primarily for dev environment, where the root is considered to be a database, though having the type of a directory. But with the present logic this change is only reasonable, since it assumes this kind of situation, when `isDatabase` conflicts with `props.type`.

Raubzeug (approved):

PR #75, opened 2022-06-23
Up types coverage and fix ColumnTables displaying

ariser:
`OlapTable` and `OlapStore` were at some point renamed to `ColumnTable` and `ColumnStore` respectively. This change was never supported in UI. This PR fixes it.

Also, the initial intent for this PR — this PR adds types coverage for the api call `json/describe`.

### DIFF
--- a/src/containers/Tenant/Tenant.tsx
+++ b/src/containers/Tenant/Tenant.tsx
@@ -28,8 +28,8 @@ import './Tenant.scss';
 const b = cn('tenant-page');
 
 export const TABLE_TYPE = 'Table';
-export const OLAP_TABLE_TYPE = 'OlapTable';
-export const OLAP_STORE_TYPE = 'OlapStore';
+export const OLAP_TABLE_TYPE = 'ColumnTable';
+export const OLAP_STORE_TYPE = 'ColumnStore';
 
 export function calcEntityType(currentPathType?: string) {
     return currentPathType && currentPathType.replace('EPathType', '');
--- a/src/containers/Tenant/utils/schema.ts
+++ b/src/containers/Tenant/utils/schema.ts
@@ -1,17 +1,20 @@
-import type {NavigationTreeNodeType} from "ydb-ui-components";
+import type {NavigationTreeNodeType} from 'ydb-ui-components';
+import {EPathType} from '../../../types/api/schema';
 
-const DB_TYPES = new Set(['EPathTypeSubDomain']);
-const TABLE_TYPES = new Set(['EPathTypeTable', 'EPathTypeOlapTable']);
-const DIR_TYPES = new Set(['EPathTypeDir', 'EPathTypeOlapStore']);
-
-export const calcNavigationTreeType = (type: string): NavigationTreeNodeType => {
-    if (DIR_TYPES.has(type)) {
-        return 'directory';
-    } else if (TABLE_TYPES.has(type)) {
-        return 'table';
-    } else if (DB_TYPES.has(type)) {
-        return 'database';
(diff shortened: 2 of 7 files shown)


src/containers/Tenant/utils/schema.ts
```diff
@@ -1,10 +1,23 @@
...
+]);
+
+export const calcNavigationTreeType = (type?: EPathType): NavigationTreeNodeType => {
+    if (!type) {
+        return 'directory';
+    }
 
-export const calcNavigationTreeType = (type: string): NavigationTreeNodeType => {
     if (DIR_TYPES.has(type)) {
         return 'directory';
```
Raubzeug, line 22: It seems it would be useful to add types also for returning value to improve extensibility.

src/containers/Tenant/utils/schema.ts
```diff
@@ -1,10 +1,23 @@
...
+    EPathType.EPathTypeColumnStore,
+]);
+
+export const calcNavigationTreeType = (type?: EPathType): NavigationTreeNodeType => {
+    if (!type) {
+        return 'directory';
+    }
 
-export const calcNavigationTreeType = (type: string): NavigationTreeNodeType => {
     if (DIR_TYPES.has(type)) {
```
Raubzeug, line 21: There are not much types, so I think it would be more clear to use `switch ... case` instead of sets.

src/containers/Tenant/Tenant.tsx
```diff
@@ -28,8 +28,8 @@ import './Tenant.scss';
 const b = cn('tenant-page');
 
 export const TABLE_TYPE = 'Table';
-export const OLAP_TABLE_TYPE = 'OlapTable';
-export const OLAP_STORE_TYPE = 'OlapStore';
+export const OLAP_TABLE_TYPE = 'ColumnTable';
+export const OLAP_STORE_TYPE = 'ColumnStore';
```
Raubzeug, line 32: Maybe we can use enum here?

Raubzeug (changes requested):

src/containers/Tenant/Tenant.tsx
ariser: This just fixes the bug. I don't want to invest more time untangling this code rn.

Raubzeug: ok, got it!

src/containers/Tenant/utils/schema.ts
ariser: Why?
The sets names are clear, and adding new types only requires extending a set

src/containers/Tenant/utils/schema.ts
ariser: The return value type is determined by the lib type `NavigationTreeNodeType`, and it is a union of strings, I don't think adding our own type would help. By the contrary, this would add another entity to modify in case there are updates in the lib.

Actually it is maybe better to rename the function to something like `mapPathTypeToNavigationTreeType`. I'll think about that.

src/containers/Tenant/utils/schema.ts
ariser: I'll modify this function a bit though, better have a default value than checking `type` for existence

ariser: I see you point though
`switch...case` also requires only one modification to extend the types
my concern is the amount of types will grow and the `switch...case` will become unreadable
I'm not sure whether this would happen though

Raubzeug (dismissed):

Raubzeug (approved):

PR #81, opened 2022-07-05
Tenant: show proper access error

ariser:
Currently 403 errors in api calls on tenant pages only show a Toast in the screen corner, keeping the page empty and the loaders spinning. This PR adds a proper Access Denied page for this situation. It displays when either `/viewer/json/tenantinfo` or `/vierer/json/describe` responds with 403.


Raubzeug (changes requested):
If we show Access Denied illustration, it is no need to show toast. It seems it should be deleted in respective cases.

ariser:
> If we show Access Denied illustration, it is no need to show toast. It seems it should be deleted in respective cases.

Actually I realised there is no toast for the 403 error. Just a blank page with an infinite spinner. Toasts are shown for http errors 500+: https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/store/utils.js#L35

@Raubzeug I think there is no such changes required as for now.

Raubzeug (approved):

PR #102, opened 2022-07-19
fix(TabletsOverall): properly hide tooltip on mouseleave

ariser:


src/containers/ReduxTooltip/ReduxTooltip.js
```diff
@@ -11,7 +11,7 @@ const propTypes = {
     className: PropTypes.string,
     toolTipVisible: PropTypes.bool,
     currentHoveredRef: PropTypes.object,
-    data: PropTypes.object,
+    data: PropTypes.any,
```
ariser, line 14: There are several types of tooltips, some of them expect data of type other than object

tsufiev (approved):

PR #106, opened 2022-07-20
feat(Node): more informative pdisks panels

ariser:
This PR introduces 2 changes:
- change data on PDisks panels
- move node basic info from the overview tab to every tab above the tabs row

Data on the PDisks panels contained Node ID, which repeated, and was redundant. Moving the overview header above the tabs makes it possible to remove Node ID from the PDisks panel while keeping this info on the page.

PDisks now contain new info about its type (SSD, NVME, etc.), which is calculated from its Category. A new field is added to the expanded PDisk view, as well as to the collapsed panel. Some other info from the expanded view now also appears on the collapsed panel.

### DIFF
--- a/src/components/BasicNodeViewer/BasicNodeViewer.scss
+++ b/src/components/BasicNodeViewer/BasicNodeViewer.scss
@@ -1,0 +1,43 @@
+@import '../../styles/mixins.scss';
+
+.basic-node-viewer {
+    font-size: var(--yc-text-body-2-font-size);
+    line-height: var(--yc-text-body-2-line-height);
+
+    display: flex;
+    align-items: center;
+
+    margin: 15px 0;
+
+    &__title {
+        margin: 0 20px 0 0;
+
+        font-size: var(--yc-text-body-2-font-size);
+        font-weight: 600;
+        line-height: var(--yc-text-body-2-line-height);
+        text-transform: uppercase;
+    }
+
+    &__id {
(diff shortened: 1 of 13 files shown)


src/utils/pdisk.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Parses a binary string containing a bit field into an object with binary values.
+ * This is an implementation based on string manipulation, since JS can only perform
+ * bitwise operations with 32-bits integers, and YDB sends uint64.
+ * @see https://en.cppreference.com/w/c/language/bit_field
+ * @param binaryString - binary string representing a bit field
+ * @param bitFieldStruct - bit field description, <field => size in bits>, in order starting from the rightmost bit
+ * @returns object with binary values
+ */
+export const parseBitField = <T extends Record<string, number>>(
```
ariser, line 13: thinking about publishing this as an npm package, haven't found anything that supports C bit fields parsing

src/components/BasicNodeViewer/BasicNodeViewer.scss
```diff
@@ -0,0 +1,43 @@
+@import '../../styles/mixins.scss';
+
+.basic-node-viewer {
```
ariser, line 3: extracted from full-node-viewer without any changes in styles, only in classes structure

Raubzeug (approved):

PR #117, opened 2022-08-03
TopShards: add DataSize column & perform sorting on backend

ariser:


src/services/api.js
```diff
@@ -147,14 +147,19 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
...
+        return this.post(
+            this.getPath('/viewer/json/query'),
+            {
+                query,
+                database,
+                action,
+                stats,
+                timeout: 600000,
+            },
+            null,
```
Raubzeug, line 160: could you please clarify, what for is this parameter?

src/containers/Tenant/Diagnostics/TopShards/TopShards.js
```diff
@@ -32,6 +42,28 @@ function prepareCPUWorkloadValue(value) {
...
+    return value && value.split(',').map((columnId) => ({
+        columnId,
+        order: 'DESC',
+    }));
+}
+
+function dataTableToStringSortOrder(value = []) {
+    return value.map(({columnId}) => columnId).join(',');
+}
+
```
Raubzeug, line 66: why do we hardcode only descending sort order?

Raubzeug (changes requested):

src/services/api.js
ariser: that's the query params
there is the method signature:
https://github.com/yandex-cloud/axios-wrapper/blob/main/src/index.ts#L231-L236

src/containers/Tenant/Diagnostics/TopShards/TopShards.js
ariser: Ascending order doesn't make sense for the top shards table. The use case is to see the top.
Moreover, if you reverse the sorting of a table which shows 20 items out of 20+ items, the resulting table would contain a different set of rows, which is confusing. One of the solutions is to enumerate rows to show that in ascending order table shows the last 20, and they are different from the first 20. The other one is to disable ascending order, which we decided to do.

ariser: there is a comment about this below, for the sort handler
I was in doubt where to put it, maybe here would be better?
although keep in mind that code navigation while reviewing differs from the code navigation when writing or reading code in ide

Raubzeug (approved):

PR #123, opened 2022-08-09
fix(QueryEditor): save chosen run action

ariser:


src/store/reducers/executeQuery.js
```diff
@@ -39,7 +39,7 @@ const initialState = {
                 ? MAXIMUM_QUERIES_IN_HISTORY - 1
                 : queriesHistoryInitial.length - 1,
     },
-    runAction: RUN_ACTIONS_VALUES.script,
+    runAction: undefined,
```
Raubzeug, line 42: I seems that `runAction` should be taken from userSettings on application initialisation.

src/containers/Tenant/QueryEditor/QueryEditor.js
```diff
@@ -650,7 +660,12 @@ function QueryEditor(props) {
 
 const mapStateToProps = (state) => {
     return {
-        executeQuery: state.executeQuery,
+        executeQuery: {
+            ...state.executeQuery,
+            runAction: state.executeQuery.runAction ||
```
Raubzeug, line 665: I think it will be more clear if `runAction` selection logic would be placed in redux store and here it should be only retrieved.

Raubzeug (changes requested):

src/store/reducers/settings.js
```diff
@@ -16,6 +23,9 @@ const savedQueries = window.web_version
 const savedTenantGeneralTab = window.web_version
     ? userSettings[TENANT_INITIAL_TAB_KEY]
     : getValueFromLS(TENANT_INITIAL_TAB_KEY);
+export const queryInitialRunAction = window.web_version
```
Raubzeug, line 26: Here is a lot of initial settings by now. Maybe it would be useful if we create a helper function that retrieves setting value depends on `window.web_version`. And in `initialState` values will be this helper calls without many constants above?

Raubzeug (changes requested):

Raubzeug (approved):

PR #128, opened 2022-08-15
feat(Storage): display donor disks

ariser:


src/containers/Storage/StorageGroups/StorageGroups.scss
```diff
@@ -1,4 +1,8 @@
 .global-storage-groups {
+    &__vdisks-column {
+        overflow: visible; // to enable stacked disks overflow the row
```
ariser, line 3: for myself: questionable, revise later

ariser: Increased stack layers offset from 3ps to 4ps, now 3-items stack will also overflow the row. Need to keep this, although 2 donors are rare.

Raubzeug (approved):

PR #129, opened 2022-08-16
Add & cleanup data in tables overview

ariser:
This PR introduces 2 changes:
1. Reformat & group table information in overview for better readability (4th commit)
2. Display new info about table read-only replicas in a new section of overview (5th commit)

the first 3 commits are technical updates to support the 4th one

### DIFF
--- a/src/components/InfoViewer/InfoViewer.scss
+++ b/src/components/InfoViewer/InfoViewer.scss
@@ -27,11 +27,10 @@
 
     &__label {
         display: flex;
-        flex: 1 1 auto;
+        flex: 0 1 auto;
         align-items: baseline;
 
         min-width: 200px;
-        max-width: 200px;
 
         white-space: nowrap;
 
--- a/src/containers/Tenant/Schema/SchemaInfoViewer/SchemaInfoViewer.scss
+++ b/src/containers/Tenant/Schema/SchemaInfoViewer/SchemaInfoViewer.scss
@@ -1,6 +1,24 @@
 .schema-info-viewer {
     overflow: auto;
 
+    &__row {
+        display: flex;
+        flex-wrap: wrap;
+        justify-content: flex-start;
+        align-items: flex-start;
+    }
+
+    &__col {
+        display: flex;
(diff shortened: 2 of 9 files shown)


src/containers/Tenant/Schema/SchemaInfoViewer/SchemaInfoViewer.scss
```diff
@@ -1,6 +1,21 @@
 .schema-info-viewer {
     overflow: auto;
 
+    &__row {
+        display: flex;
+        flex-wrap: wrap;
+        justify-content: flex-start;
+        align-items: flex-start;
+        gap: 50px;
```
Raubzeug, line 9: Flex gap is not supported id 14 Safari

Raubzeug (changes requested):

src/containers/Tenant/Schema/SchemaInfoViewer/SchemaInfoViewer.scss
ariser: And still it is used all over the project. We probably should address that, if we target safari 14.
Are you suggesting we don't use it here?

Raubzeug (dismissed):

src/services/api.js
```diff
@@ -83,7 +83,6 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
                 path,
                 enums: true,
                 backup: false,
-                partition_config: false,
```
ariser, line 86: partition data didn't load, fixed

src/components/InfoViewer/InfoViewer.scss
```diff
@@ -27,11 +27,10 @@
 
     &__label {
         display: flex;
-        flex: 1 1 auto;
+        flex: 0 1 auto;
         align-items: baseline;
 
         min-width: 200px;
-        max-width: 200px;
```
ariser, line 34: to allow longer labels widen the block, like this

Raubzeug (approved):

PR #138, opened 2022-08-25
fix(Storage): expect arbitrary donors data

ariser:
Fix for crashes caused by unexpected data in the Donors field.
Along with that, more correct types for some fields: uint32 via HTTP received as number, not string.

### DIFF
--- a/src/containers/Storage/utils/index.ts
+++ b/src/containers/Storage/utils/index.ts
@@ -1,1 +1,6 @@
+import type {TVDiskStateInfo, TVSlotId} from '../../../types/api/storage';
+
 export * from './constants';
+
+export const isFullDonorData = (donor: TVDiskStateInfo | TVSlotId): donor is TVDiskStateInfo =>
+    'VDiskId' in donor;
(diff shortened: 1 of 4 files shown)


src/containers/Storage/utils/index.ts
```diff
@@ -1 +1,6 @@
+import type {TVDiskStateInfo} from '../../../types/api/storage';
+
 export * from './constants';
+
+export const isFullDonorsData = (donors: TVDiskStateInfo['Donors']): donors is TVDiskStateInfo[] =>
+    Array.isArray(donors) && donors.length > 0 && 'VDiskId' in donors[0];
```
Raubzeug, line 6: Is this enough to check if `VDiskId` is defined only in the first donor?

Raubzeug (changes requested):

src/containers/Storage/utils/index.ts
ariser: The array is either all `TVDiskStateInfo`, or all `TVSlotId`. But I suppose it would be more safe to define the field as `Array<TVDiskStateInfo | TVSlotId>` instead of `TVDiskStateInfo[] | TVSlotId[]`, and check every item. No harm in doing this, I'll update it, thanks.

Raubzeug (approved):

PR #139, opened 2022-08-25
feat(Storage): show usage column

ariser:


src/containers/Storage/StorageGroups/StorageGroups.tsx
```diff
@@ -120,6 +120,24 @@ function StorageGroups({data, tableSettings, visibleEntities, nodes}: StorageGro
...
+            header: tableColumnsNames[TableColumnsIds.UsedPercents],
+            width: 100,
+            render: ({row}) => {
+                const usage = getUsage(row, 5);
+                return (
+                    <Label
+                        theme={getUsageSeverity(usage)}
+                        className={b('usage-label', {overload: usage >= 100})}
+                    >
+                        ≥ {usage}%
```
Raubzeug, line 134: It seems it looks strange. Just as example: my calculated usage equals to 20%. And I see on screen >=20%. So... is it 21 or 99 percent?
Maybe `~` symbol fits better?

src/containers/Storage/utils/index.ts
```diff
@@ -40,3 +40,11 @@ export const getDegradedSeverity = (group: IStoragePoolGroup) => {
 
     return evaluate(group.Missing);
 };
+
+export const getUsageSeverity = generateEvaluator(80, 85);
+
+export const getUsage = (data: IStoragePoolGroup, step = 1) => {
```
Raubzeug, line 46: should check if step is not equals zero.

src/containers/Storage/StorageGroups/StorageGroups.tsx
```diff
@@ -120,6 +120,24 @@ function StorageGroups({data, tableSettings, visibleEntities, nodes}: StorageGro
...
+        {
+            name: TableColumnsIds.UsedPercents,
+            header: tableColumnsNames[TableColumnsIds.UsedPercents],
+            width: 100,
+            render: ({row}) => {
+                const usage = getUsage(row, 5);
+                return (
+                    <Label
+                        theme={getUsageSeverity(usage)}
+                        className={b('usage-label', {overload: usage >= 100})}
```
Raubzeug, line 132: I suggest to discuss edge value - maybe 95% is already critical?

Raubzeug (changes requested):

src/containers/Storage/StorageGroups/StorageGroups.tsx
ariser: Good catch. I thought about it. It will be clear in comparison to other rows. But in case with few rows this is still an issue.

This is how it is shown in design though. I would keep it this way for now and discussed with the team once again.

src/containers/Storage/StorageGroups/StorageGroups.tsx
ariser: Critical is 85, and is colored red. 100 and greater is just redder :)

ariser: But I add this to the questions that need discussing regarding this feature, thanks

src/containers/Storage/utils/index.ts
ariser: I'm not convinced we should. With step=0 it just will result in NaN, which is correct. This argument is code-dependent, not data-dependent, so there is no chance of error because of wrong data, only because of a programmer’s error, and I don’t think checking will make it better somehow, since zero step doesn’t make sense anyway, and NaN is the correct result for it.

Raubzeug (approved):

PR #146, opened 2022-08-30
feat(Storage): usage filter

ariser:


src/containers/Storage/UsageFilter/UsageFilter.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const handleUpdate = (newValue: string[]) => {
+        setFilterValue(newValue);
+
+        window.clearTimeout(timer.current);
+        timer.current = window.setTimeout(() => {
+            onChange?.(newValue);
+        }, debounce);
+    };
+
+    const maxWidth = Math.max(...groups.map(({count}) => count));
```
Raubzeug, line 68: What if `groups` is empty array?

Raubzeug (changes requested):

src/containers/Storage/UsageFilter/UsageFilter.tsx
ariser: Then `maxWith` is `-Inifinity`, and it won't be used, since `groups` is an empty array, and `maxWidth` is only used in rendering an option

Raubzeug (approved):

PR #150, opened 2022-09-01
fix(Tenant): move nodes tab from diagnostics to top level

ariser:


Raubzeug (approved):

ariser:
Seems like this change doesn't solve the original problem, which is to disambiguate similar tabs: Nodes in the AsideNav, and Nodes in the database context. Postponing merge until we agree on the solution.

PR #154, opened 2022-09-03
fix(Tenant): load root if cahced path is not in tree

ariser:
In SPA mode, when switching between adjacent databases in the same cluster, `currentSchamePath` from the previous database preserves, and tries to load. When it exists (e.g. when opening an ancestor path), everything is fine. When it does not exist, nothing is selected in the tree, and nothing is loaded.

This PR fixes this case and shows the root node when the cashed path does not exist in the tree — assuming every path in a tree starts with the root path.

### DIFF
--- a/src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
+++ b/src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
@@ -1,3 +1,4 @@
+import {useEffect} from 'react';
 import {useDispatch} from 'react-redux';
 
 import {NavigationTree} from 'ydb-ui-components';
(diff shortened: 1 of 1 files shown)


src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
```diff
@@ -48,6 +49,13 @@ export function SchemaTree(props: SchemaTreeProps) {
         dispatch(getSchemaAcl({path: activePath}));
     };
 
+    useEffect(() => {
+        // if the cached path is not in the current tree, show root
+        if (!currentPath.startsWith(rootPath)) {
+            handleActivePathUpdate(rootPath);
+        }
+    }, []);
```
Raubzeug, line 57: If it would be better to clean cache in cleanup function in useEffect? If we changed database - doesn't it mean that old caches are not useful any more?

Raubzeug (changes requested):

src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
ariser: We may return to the same database, then it is ok to show the previously open node. With clearing on unmount it will always be root — and this is a change of the current behavior, which I don't think I want to do.

Raubzeug (approved):

PR #171, opened 2022-09-20
fix: modern query format

ariser:
This PR addresses two issues:
1. Response format of `viewer/json/query`
2. Column tables info tab data fetching

1 — after adopting the `result` field in query response in PRs #168 and #170, this finishes the change, applying the proper response parser and making sure the API is called with the proper params

2 — while updating column tables info tab to work with the new query response format, I bumped into a bug with data refetching. Column tables require extra data for the Info tab, which is fetched from `viewer/json/query`. The bug is, in some cases (more often than not) the fetch wasn't initiated upon schema navigation. This PR fixes this, and the proper data always fetch for column tables.


sareyu (approved):
lgtm

PR #188, opened 2022-10-08
fix(QueryEditor): smarter error message trim

ariser:
Don't trim errors if the toggle label is longer than the trimmed part

### DIFF
--- a/src/components/ShortyString/ShortyString.tsx
+++ b/src/components/ShortyString/ShortyString.tsx
@@ -10,6 +10,8 @@ const block = cn('kv-shorty-string');
 type Props = {
     value?: string;
     limit?: number;
+    /** in strict mode the text always trims at the limit, otherwise it is allowed to overflow a little */
+    strict?: boolean;
     displayLength?: boolean;
     render?: (value: string) => React.ReactNode;
     onToggle?: () => void;
@@ -20,6 +22,7 @@ type Props = {
 export default function ShortyString({
     value = '',
     limit = 200,
+    strict = false,
     displayLength = true,
     render = (v: string) => v,
     onToggle,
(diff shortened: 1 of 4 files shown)


src/components/ShortyString/ShortyString.tsx
```diff
@@ -27,12 +27,21 @@ export default function ShortyString({
...
-    const label = expanded ? collapseLabel : expandLabel;
+    const toggleLabelAction = expanded ? collapseLabel : expandLabel;
+    const toggleLabelSymbolsCount = displayLength && !expanded
+        ? ` (${value.length} symbols)`
+        : '';
+    const toggleLabel = toggleLabelAction + toggleLabelSymbolsCount;
+
+    // showing toogle button with a label that is longer than the hidden part is pointless,
+    // hence compare to limit + length
+    const hasToggle = value.length > limit + toggleLabel.length;
```
artemmufazalov, line 39: toggleLabel length should also include 4 additional symbols (three dots and a space, they are counted in slice function below). Besides, length of margin (2 em) between text and button could be considered here as well

src/components/ShortyString/ShortyString.tsx
```diff
@@ -27,12 +27,21 @@ export default function ShortyString({
...
+
+    const text = expanded || !hasToggle
+        ? value
```
artemmufazalov, line 42: Having set limit for the value to a certain level via props, in fact the length of a shown text will be the limit + the length of the button. For example, in the app there is a place where it will be shown more than 230 symbols instead of limit set to 200. Won’t it be a little misleading?

src/components/ShortyString/ShortyString.tsx
ariser: They are included in the limit. The final string is `text - 4 chars + space + 3 dots` which length equals to `limit`.

Considering the margin is an overkill, the point of this change is not to trim the text if the toggle label is longer than the hidden part.

src/components/ShortyString/ShortyString.tsx
ariser: A valid point. The the simplest way out of this is to introduce a prop that would toggle this behavior. Will do that.

src/components/ShortyString/ShortyString.tsx
```diff
@@ -20,19 +23,29 @@ type Props = {
...
     const [expanded, setExpanded] = React.useState(false);
-    const hasToggle = value.length > limit;
-    const length =
-        displayLength && !expanded ? `(${value.length} symbols)` : undefined;
 
-    const text = expanded || value.length <= limit ? value : value.slice(0, limit - 4) + '\u00a0...';
-    const label = expanded ? collapseLabel : expandLabel;
+    const toggleLabelAction = expanded ? collapseLabel : expandLabel;
+    const toggleLabelSymbolsCount = displayLength && !expanded
+        ? ` (${value.length} symbols)`
```
artemmufazalov, line 37: Since you add i18n translations, it will be great to translate “symbols” as well

artemmufazalov (changes requested):

src/components/ShortyString/ShortyString.tsx
sareyu: and plural form also

sareyu (approved):

src/components/ShortyString/ShortyString.tsx
ariser: missed it, thanks

artemmufazalov (approved):

PR #191, opened 2022-10-11
fix: case insensitive search

ariser:


artemmufazalov (approved):

src/containers/Nodes/Nodes.js
```diff
@@ -114,11 +114,14 @@ class Nodes extends React.Component {
...
-            ? nodes.filter((node) =>
-                  node.Host
-                      ? node.Host.includes(searchQuery) || String(node.NodeId).includes(searchQuery)
-                      : true,
-              )
+            ? nodes.filter((node) => {
+                  const lowerCaseSearchQuery = searchQuery.toLowerCase();
+
+                  return node.Host
+                      ? node.Host.toLowerCase().includes(lowerCaseSearchQuery) ||
```
sareyu, line 121: maybe it would be cleaner using case insensitive regexp?

```
const re = new RegExp(searchQuery, ig);
return re.test(node.Host) || re.test(node.NodeId); // need to check if prev match case works correctly
```

sareyu (dismissed):

src/containers/Nodes/Nodes.js
ariser: This looks cleaner, I agree.
The only correction is the global modifier shouldn't be used here, as it increments `lastIndex`.

Funny thing is, this would make the second comparison only conceptually incorrect. In this case it would work correctly though, since the first test would find nothing and keep `lasIndex` at 0 :)

artemmufazalov (approved):

PR #192, opened 2022-10-12
feat(Healthcheck): redesign healthcheck popup

artemmufazalov:


src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
```diff
@@ -77,67 +41,69 @@ const IssuesViewer = ({issues}) => {
...
-                        }
-                        collapsed={
-                            typeof collapsedIssues[id] === 'undefined' || collapsedIssues[id]
-                        }
-                        hasArrow={hasArrow}
-                        onClick={() => setActiveItem(id)}
+                        name={<IssueRow data={item} />}
+                        collapsed={isCollapsed}
+                        hasArrow={true}
+                        onClick={() => {
```
sareyu, line 68: unnecessary lambda

ariser (changes requested):

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
```diff
@@ -77,67 +41,61 @@ const IssuesViewer = ({issues}) => {
...
-                            const newValue =
-                                typeof collapsedIssues[id] === 'undefined'
-                                    ? false
-                                    : !collapsedIssues[id];
-                            const newCollapsedIssues = {...collapsedIssues, [id]: newValue};
-                            setCollapsedIssues(newCollapsedIssues);
-                        }}
+                        name={<IssueRow data={item} />}
+                        collapsed={isCollapsed}
+                        hasArrow={true}
```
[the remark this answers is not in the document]
artemmufazalov: Previously, a block has an arrow only if it has reason issues, now block also has an info panel that is expanded with this arrow. That is why it is always true

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
```diff
@@ -48,6 +48,10 @@ const IssuesViewer = ({issues}) => {
                 // eslint-disable-next-line no-unused-vars
                 const {status, message, type, reasonsItems, reason, level, ...rest} = item;
 
+                if (level === 1 && expandedIssueId && id !== expandedIssueId) {
```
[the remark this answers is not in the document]
artemmufazalov: Here we should clarify what results we could possibly get from the backend, whether it is possible to get any n-level issues that do not have their first level “parent”. 

We show only first level issues in a preview, so modal is opened for a specific first level issue. If it is possible to have "parentless" n-level issues, they will be displayed in the modal too. For this case we should rework preview component as well

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
```diff
@@ -77,67 +41,61 @@ const IssuesViewer = ({issues}) => {
...
-                const hasArrow = item[childrenKey].length;
+
+                // eslint-disable-next-line no-unused-vars
+                const {status, message, type, reasonsItems, reason, level, ...rest} = item;
+
+                const isCollapsed =
+                    typeof collapsedIssues[id] === 'undefined' || collapsedIssues[id];
+
+                const toggleCollapsed = () => {
+                    setCollapsedIssues({...collapsedIssues, [id]: !isCollapsed});
```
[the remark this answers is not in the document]
artemmufazalov: Could you please clarify, why callback is preferred in this case (and other complex enough cases)?

Is it needed to optimize rerenders, as the documentation says?

“If the new state is computed using the previous state, you can pass a function to setState. The function will receive the previous value, and return an updated value.

If your update function returns the exact same value as the current state, the subsequent rerender will be skipped completely.”

(from https://reactjs.org/docs/hooks-reference.html)

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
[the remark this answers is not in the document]
ariser: Ok, I looked into it and see now how the data is processed. Backend sends a plain list of issues, which is then transformed into a tree structure. My first comment is not applicable, the current solution works. But we need to redo it :)

The conversion from a plain list into a tree structure should be done in a selector, not in a component. We use reselect in other reducers, and should use it in `healthcheckInfo`.

Moreover, the whole components system would be better designed, if `IssuesViewer` accepted a final data structure to display and didn't have an internal filtering logic. The final structure is a tree, so all we need to do is either to pass a subtree from the parent component, or select it from the store.

Along with all this, we can as well migrate `IssuesVewer` to TS, the refactoring would already change a lot

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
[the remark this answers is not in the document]
ariser: The optimization part in the docs is just to let you know how the callback handles the unchanged state.

The point is that callback is guaranteed to have the correct previous state value as an argument, whereas the usage of a state variable can be unreliable.

A more obvious example is changing state within a `setInterval`, where the state variable would have only the initial state, therefore the only way to update the state would be via callback.
```js
const [counter, setCounter] = useState(0);
useEffect(() => {
    setInterval(() => {
        setCounter(val => val + 1); // using `counter` wouldn't work here
    }, 1000);
}, []);
```

In the current case you _can_ use the state variable, but it doesn't mean you should :) You don't know what changes will be introduced to this component in the future, it is generally a good idea to stick to a safer solution.

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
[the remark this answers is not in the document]
artemmufazalov: Thanks, it’s a good notice. I will refactor it, though it requires some more time

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssuesViewer.js
```diff
@@ -70,9 +70,10 @@ const IssuesViewer = ({issues, expandedIssueId}) => {
                         hasArrow={true}
                         onClick={toggleCollapsed}
                         onArrowClick={toggleCollapsed}
+                        level={treeLevel}
```
ariser, line 73: there is level in the data

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssueViewer.scss
```diff
@@ -172,24 +157,18 @@
...
+        // Without !important this class does not have enough weight compared to styles set in TreeView
+        .ydb-tree-view__item {
+            margin-left: $calculated-margin !important;
+            padding-left: 0 !important;
         }
 
-        .ydb-tree-view .ydb-tree-view .ydb-tree-view .ydb-tree-view {
-            @include set-tree-margins(4);
+        .issue-viewer__info-panel {
+            margin-left: $calculated-margin;
```
ariser, line 167: margins for adjacent info-panel and for __item differ by 1 level

ariser: Ah no they do not. My mistake, everything is ok here

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
```diff
@@ -31,11 +31,14 @@ export const Healthcheck = (props: HealthcheckProps) => {
...
+    const fetchHealthcheck = useCallback(
+        (isForcedFetch = false) => {
+            if (fetchData || isForcedFetch) {
+                dispatch(getHealthcheckInfo(tenant));
+            }
+        },
+        [dispatch, fetchData, tenant],
+    );
 
     useAutofetcher(fetchHealthcheck, [fetchHealthcheck], autorefresh);
```
ariser, line 43: i'd just moved the test of `fetchData` here instead of introducing `isForcedFetch`

src/containers/Tenant/Diagnostics/Healthcheck/IssuesViewer/IssueViewer.scss
```diff
@@ -9,18 +9,16 @@
     cursor: pointer;
 
     &__field {
-        padding: 0 10px;
+        display: flex;
+        justify-content: space-between;
```
ariser, line 13: this doesn't make sense here, this is a property of a flex container, and `__field` is a flex item

well, it is a container since you made it one, but they both contain a single child, so they don't need to be

ariser:
feel free to squash the fixups, I'll approve then

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
artemmufazalov: I found your solution much cleaner, so i applied it

ariser (approved):

PR #194, opened 2022-10-17
refactor: Healthcheck issues

artemmufazalov:
Refactor some of the Healthcheck components and healthcheckInfo reducer moving issues tree creation logic to reducer and rewriting them to TS.

Inspired by @ariser comment https://github.com/ydb-platform/ydb-embedded-ui/pull/192#discussion_r994013742

### DIFF
--- a/src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
+++ b/src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
@@ -6,30 +6,30 @@ import updateArrow from '../../../../../assets/icons/update-arrow.svg';
 
 import type {IIssuesTree} from '../../../../../types/store/healthcheck';
 
-import IssuesViewer from '../IssuesViewer/IssuesViewer';
+import IssueTree from '../IssuesViewer/IssueTree';
 
 import i18n from '../i18n';
 
 const b = cn('healthcheck');
 
-interface IssuesListProps {
-    issue?: IIssuesTree;
+interface DetailsProps {
+    issueTree?: IIssuesTree;
     loading?: boolean;
     onUpdate: VoidFunction;
 }
 
-export const IssuesList = (props: IssuesListProps) => {
-    const {loading, onUpdate, issue} = props;
+export const Details = (props: DetailsProps) => {
+    const {loading, onUpdate, issueTree} = props;
 
-    if (!issue) {
+    if (!issueTree) {
         return null;
     }
 
     const renderHealthcheckHeader = () => {
         return (
-            <div className={b('issues-list-header')}>
-                <h3 className={b('issues-list-header-title')}>{i18n('title.healthcheck')}</h3>
-                <div className={b('issues-list-header-update')}>
+            <div className={b('details-header')}>
+                <h3 className={b('details-header-title')}>{i18n('title.healthcheck')}</h3>
+                <div className={b('details-header-update')}>
                     <Button size="s" onClick={onUpdate} loading={loading} view="flat-secondary">
                         <Icon data={updateArrow} height={20} width={20} />
                     </Button>
@@ -40,14 +40,14 @@ export const IssuesList = (props: IssuesListProps) => {
 
     const renderHealthcheckIssues = () => {
         return (
-            <div className={b('issues')}>
-                <IssuesViewer issue={issue} />
+            <div className={b('issues-wrapper')}>
+                <IssueTree issueTree={issueTree} />
             </div>
         );
     };
 
     return (
-        <div className={b('issues-list')}>
+        <div className={b('details')}>
             {renderHealthcheckHeader()}
             {renderHealthcheckIssues()}
         </div>
--- a/src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
+++ b/src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
@@ -7,8 +7,8 @@ import {Loader} from '@gravity-ui/uikit';
 import {SelfCheckResult} from '../../../../types/api/healthcheck';
 import {
     getHealthcheckInfo,
-    selectIssuesTrees,
-    selectIssueTreeById,
+    selectIssuesTreeById,
+    selectIssuesTreesRoots,
 } from '../../../../store/reducers/healthcheckInfo';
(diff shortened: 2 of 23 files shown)


src/store/reducers/healthcheckInfo.ts
```diff
@@ -0,0 +1,122 @@
+import _ from 'lodash';
```
ariser, line 1: I'd get rid of lodash where it is possible
But more importantly I'd get rid of these kinds of lodash imports. The package supports tree-shaking by using explicit imports like the ones below. This one imports the whole lib.

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
```diff
@@ -61,15 +73,15 @@ export const Healthcheck = (props: HealthcheckProps) => {
         if (data && data['self_check_result']) {
             return preview ? (
                 <Preview
-                    data={data}
+                    issuesTrees={issuesTrees}
```
ariser, line 76: Is there a point in passing the whole trees there? Only the roots are displayed.
There also may be no point in creating a selector to only select the roots. This is a question to consider and prove, not to blindly fix.

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
```diff
@@ -61,15 +73,15 @@ export const Healthcheck = (props: HealthcheckProps) => {
...
+                    selfCheckResult={selfCheckResult}
                     loading={loading}
                     onShowMore={showMoreHandler}
                     onUpdate={fetchHealthcheck}
                 />
             ) : (
                 <IssuesList
-                    data={data}
-                    expandedIssueId={expandedIssueId}
+                    issue={expandedIssueConsequence}
```
ariser, line 84: every item in an issues tree is an issue, so I'd say this naming is misleading

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
```diff
@@ -4,7 +4,12 @@ import cn from 'bem-cn-lite';
...
-import {getHealthcheckInfo} from '../../../../store/reducers/healthcheckInfo';
+import {SelfCheckResult} from '../../../../types/api/healthcheck';
+import {
+    getHealthcheckInfo,
+    selectInvertedIssuesConsequenceTrees,
+    selectIssueConsequenceById,
+} from '../../../../store/reducers/healthcheckInfo';
 import {useAutofetcher} from '../../../../utils/hooks';
 
 import {IssuesList} from './IssuesList';
```
ariser, line 15: as for the naming, i'd go with something like

```
Healthcheck/
  Preview
  Details // currently IssuesList 
  IssuesTree // currently IssuesViewer
```

`IssuePreview` is clashing with `Preview`. So there may be point in moving it inside Preview, like `IssueRow` is inside `IssuesViewer` now. On the other hand, if `IssuePreview` remains a separate component, we should make an `IssueRow` one as well, to maintain structural consistency.

In case they are separate components, they could be named something like `PreviewItem` and `IssuesTreeItem`

ariser (changes requested):

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
artemmufazalov: I think it makes sense since trees can be big. Also in this case we work with the whole trees only in reducer while components get either roots or single tree

src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
```diff
@@ -6,30 +6,30 @@ import updateArrow from '../../../../../assets/icons/update-arrow.svg';
...
-import IssuesViewer from '../IssuesViewer/IssuesViewer';
+import IssueTree from '../IssuesViewer/IssueTree';
 
 import i18n from '../i18n';
 
 const b = cn('healthcheck');
 
-interface IssuesListProps {
-    issue?: IIssuesTree;
+interface IssuesDetailsProps {
```
ariser, line 15: that's healthcheck details, not issues details
and usually it is based on the component naming, so `DetailsProps`

src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
```diff
@@ -6,30 +6,30 @@ import updateArrow from '../../../../../assets/icons/update-arrow.svg';
...
+    if (!issueTree) {
         return null;
     }
 
     const renderHealthcheckHeader = () => {
         return (
-            <div className={b('issues-list-header')}>
-                <h3 className={b('issues-list-header-title')}>{i18n('title.healthcheck')}</h3>
-                <div className={b('issues-list-header-update')}>
+            <div className={b('issues-details-header')}>
```
ariser, line 30: same for the class names

src/store/reducers/healthcheckInfo.ts
```diff
@@ -104,6 +105,10 @@ export const selectIssueTreeById = createSelector(
     (issueTrees, id) => issueTrees.find((issueTree: IIssuesTree) => issueTree.id === id),
 );
 
+export const selectIssuesTreesRoots = createSelector(selectIssuesTrees, (issuesTrees) =>
+    issuesTrees.map((issueTree) => _omit(issueTree, ['reasonsItems'])),
+);
```
ariser, line 110: is this more efficient than just filtering the raw plain response?

ariser (changes requested):

ariser (changes requested):

src/store/reducers/healthcheckInfo.ts
```diff
@@ -95,7 +95,7 @@ const getInvertedConsequencesTree = ({
         : [];
 };
 
-const getIssuesLog = (state: any) => state.healthcheckInfo.data?.issue_log as IssueLog[];
+const getIssuesLog = (state: any): IssueLog[] => state.healthcheckInfo.data?.issue_log;
```
ariser, line 98: `| undefined`

ariser (approved):

PR #196, opened 2022-10-18
fix(Tenant): show loader when fetching overview data

ariser:
This fixes the bug with a flash of an empty state when navigating a tenant schema.

a few reasons for this bug:
1. `wasLoaded` didn't reset on active path change. Because of this loader didn't display, as it should for background updates — fixed by setting `wasLoaded` to false on active path change
2. because of too many requests for `viewer/json/describe`, some of them were being cancelled by newer ones. A request cancellation is treated as a request error, which sets `loading` to false. Fixed by ignoring errors caused by requests cancellation and not changing the store for them.
3. another consequence of too many parallel requests to `viewer/json/describe` is a more complex bug, which is still present. `Tenant` initiates 2 requests for its root and the selected node on active path change. They both use the same reducer and the same actions. The request for the root is handled incorrectly because of a faulty reducer logic, clearing data for the current node. Requests for the current node resolve later, and because of that data coincidently displays correctly. But the flash of empty state still occurs. This is not fixed, but there is a workaround that resets `wasLoaded` to false before every of these requests, making loader display until all of the requests are resolved.

The 3rd reason requires more attention. The requests to `viewer/json/describe` require refactoring and optimization. Some of them are excessive.

### DIFF
--- a/src/store/reducers/olapStats.js
+++ b/src/store/reducers/olapStats.js
@@ -6,6 +6,7 @@ import {createRequestActionTypes, createApiRequest} from '../utils';
 
 const FETCH_OLAP_STATS = createRequestActionTypes('query', 'SEND_OLAP_STATS_QUERY');
 const SET_OLAP_STATS_OPTIONS = createRequestActionTypes('query', 'SET_OLAP_STATS_OPTIONS');
+const RESET_LOADING_STATE = 'olapStats/RESET_LOADING_STATE';
 
 const initialState = {
     loading: false,
@@ -48,6 +49,12 @@ const olapStats = (state = initialState, action) => {
                 ...state,
                 ...action.data,
             };
+        case RESET_LOADING_STATE: {
+            return {
+                ...state,
+                wasLoaded: initialState.wasLoaded,
+            };
+        }
         default:
             return state;
     }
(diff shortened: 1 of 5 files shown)


src/utils/hooks/useAutofetcher.ts
```diff
@@ -2,7 +2,11 @@ import {DependencyList, useEffect, useRef} from 'react';
 
 import {AutoFetcher} from '../autofetcher';
 
-export const useAutofetcher = (fetchData: VoidFunction, deps: DependencyList, enabled = true) => {
+export const useAutofetcher = (
+    fetchData: (isBackground: boolean) => void,
```
artemmufazalov, line 6: In Healthcheck we pass to autoFetcher fetchData without any params.

Is it true, that for TS it is valid to pass function without arguments when function with arguments is expected? Will it be OK to pass isBackground flag to function that does not expect it?

src/store/reducers/olapStats.js
```diff
@@ -73,4 +80,10 @@ export function setOlapStatsOptions(options) {
     };
 }
 
+export function resetLoadingState() {
+    return {
+        type: RESET_LOADING_STATE,
```
sareyu, line 85: why do you use two action with exact same type? one is enough

src/utils/hooks/useAutofetcher.ts
ariser: Yea, it's fine. This signature makes sure the call expression passes the required arguments into the function. It is not mandatory for the function to accept and use them.

src/store/reducers/olapStats.js
ariser: they are for two different reducers, types are differ by prefixes

artemmufazalov (approved):

PR #199, opened 2022-10-19
feat(Storage): display disks and groups media types

ariser:


src/containers/Storage/StorageGroups/StorageGroups.tsx
```diff
@@ -119,7 +121,19 @@ function StorageGroups({data, tableSettings, visibleEntities, nodes, onShowAll}:
         {
             name: TableColumnsIds.Type,
             header: tableColumnsNames[TableColumnsIds.Type],
-            render: ({value}) => <Label>{value as string || '—'}</Label>,
+            render: ({value, row}) => (
+                <>
+                    <Label>{value as string || '—'}</Label>
+                    {' '}
+                    {row.Encryption && (
+                        <Tooltip content={i18n('encrypted')} placement="right">
```
artemmufazalov, line 129: It seems like here we have cursor pointer for elements with tooltips.

So either we need to add such a cursor here, or set default cursor for Pool Name or FDQN (in Nodes), which have tooltips, but are not clickable.

Also the size of the tooltip differs from the size of PoolName tooltip, as well as differs tooltip color in light theme.

artemmufazalov (changes requested):

src/containers/Storage/StorageGroups/StorageGroups.tsx
ariser: these are all good points, and there is an incoming PR with these changes :) I don't want to mix refactoring with a new feature which is this PR

cursor: pointer should indicate that the element is clickable, not that is has a tooltip
I believe pool name and fqdn have it because of `Popover` that is used to show tooltips. I'm changing it to `Tooltip` in the next PR, it will also change cursor to default, and equalize the tooltip styles

src/containers/Storage/DiskStateProgressBar/DiskStateProgressBar.tsx
```diff
@@ -63,13 +59,7 @@ function DiskStateProgressBar({
...
             aria-valuenow={diskAllocatedPercent}
         >
-            {href ? (
-                <InternalLink to={href} className={b('link')}>
-                    {renderAllocatedPercent()}
-                </InternalLink>
-            ) : (
-                renderAllocatedPercent()
-            )}
+            {renderAllocatedPercent()}
```
ariser, line 62: This makes this component a single purpose, only displaying a disk state. It can be wrapped in a link if needed. Simultaneously, this fixes the issue with media type label blocking the link active area

src/containers/Storage/Pdisk/Pdisk.scss
```diff
@@ -1,18 +1,12 @@
 .pdisk-storage {
     position: relative;
 
-    display: flex;
-    flex-grow: 1;
-    align-items: center;
-
     min-width: 120px;
-    max-width: 200px;
-    margin-right: 10px;
```
ariser: These styles make no sense for this component. They belong to the parent component which is responsible for displaying a group of disks in a row. I made a similar change for vdisks previously: https://github.com/ydb-platform/ydb-embedded-ui/commit/87be45db90e381ac4ff98ccd159e5d503d4d50c7
now pdisks match them.

src/containers/Storage/StorageNodes/StorageNodes.scss
```diff
@@ -7,6 +7,16 @@
 
         min-width: 500px;
     }
+    &__pdisks-item {
```
ariser, line 10: now this handles the layout of disks groups

src/containers/Storage/Vdisk/Vdisk.js
```diff
@@ -225,33 +230,41 @@ function Vdisk(props) {
...
-            return;
+            return undefined;
         }
-        return !isNaN(Number(AllocatedSize))
-            ? (Number(AllocatedSize) * 100) / (Number(available) + Number(AllocatedSize))
-            : undefined;
+
+        return isNaN(Number(AllocatedSize))
+            ? undefined
+            : (Number(AllocatedSize) * 100) / (Number(available) + Number(AllocatedSize));
```
ariser, line 238: changes to satisfy the linter: fix inconsistent return type and a negated condition

src/containers/Storage/StorageGroups/StorageGroups.tsx
ariser: Changed my mind. Will use Popover here instead of Tooltip. They differ in behaviour, e.g. Popover allows its content to be selected, whereas Tooltip hides when cursor leaves the anchor element. I'd rather change other use cases to Tooltip, it looks like a better fit. But the selectability may be an important feature. So it's better for the new case to match the old ones.

src/containers/Storage/Pdisk/Pdisk.scss
```diff
@@ -1,18 +1,25 @@
...
-    flex-grow: 1;
-    align-items: center;
+    position: relative;
 
-    max-width: 200px;
-    margin-right: 10px;
+    min-width: 120px;
 
-    cursor: pointer;
+    border-radius: 4px; // to match interavtive area with disk shape
```
artemmufazalov, line 6: typo (interaVtive)

src/containers/Storage/Pdisk/Pdisk.scss
```diff
@@ -1,18 +1,25 @@
...
 
-    &:last-child {
-        margin-right: 0px;
+    &__content {
+        border-radius: 4px; // to match interavtive area with disk shape
```
artemmufazalov, line 9: typo (interaVtive)

src/containers/Storage/Vdisk/Vdisk.scss
```diff
@@ -1,4 +1,10 @@
 .vdisk-storage {
+    border-radius: 4px; // to match interavtive area with disk shape
```
artemmufazalov, line 2: typo (interaVtive)

src/containers/Storage/Vdisk/Vdisk.scss
```diff
@@ -1,4 +1,10 @@
...
+
+    &__content {
+        border-radius: 4px; // to match interavtive area with disk shape
```
artemmufazalov, line 5: typo (interaVtive)

artemmufazalov (changes requested):

artemmufazalov (dismissed):

artemmufazalov (approved):

PR #200, opened 2022-10-19
fix(Storage): shrink tooltip active area on FQDN

ariser:
This PR is miscellaneous changes
- fixes tooltip layout for the Nodes section of Storage the same way it was fixed for the Groups section in #136 
- introduces a css variable for table row hover color, making maintenance easier
- does a bit of cleanup for Storage tooltips

### DIFF
--- a/src/containers/App/App.scss
+++ b/src/containers/App/App.scss
@@ -31,6 +31,10 @@ body,
     --data-table-row-height: 40px;
 }
 
+.yc-root {
+    --ydb-data-table-color-hover: var(--yc-color-base-float-hover);
+}
+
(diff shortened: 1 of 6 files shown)


src/containers/App/App.scss
```diff
@@ -130,7 +130,7 @@ body,
 }
 
 .yc-root .data-table_highlight-rows .data-table__row:hover {
-    background: var(--yc-color-base-float-hover);
+    background: var(--yc-color-base-simple-hover);
```
ariser, line 133: better introduce a new var for this purpose

artemmufazalov (approved):

PR #201, opened 2022-10-20
fix(Diagnostics): fix tabs reset on page reload

artemmufazalov:
On page reload path type is undefined, so default values are set for tabs:
`const DIR_PAGES = [overview, topShards, describe];`

If there is not tab from search query in this list, the tab is set to `overview`, so it is impossible to get for example TopShard or Tablets tab right after page reload.

To fix it wasLoaded flag for tenant reducer is introduced, so the tab is set only after content is initially loaded. While the content is loading, tabs are hidden behind Loader.

### DIFF
--- a/src/containers/Tenant/Diagnostics/Diagnostics.scss
+++ b/src/containers/Tenant/Diagnostics/Diagnostics.scss
@@ -5,6 +5,13 @@
 
     height: 100%;
 
+    &__loader {
+        display: flex;
+        flex-grow: 1;
+        justify-content: center;
+        align-items: center;
+    }
+
     &__header-wrapper {
         padding: 13px 20px 16px;
 
--- a/src/containers/Tenant/Diagnostics/Diagnostics.tsx
+++ b/src/containers/Tenant/Diagnostics/Diagnostics.tsx
@@ -5,7 +5,7 @@ import {Link} from 'react-router-dom';
 import {useDispatch, useSelector} from 'react-redux';
 import {useLocation} from 'react-router';
 
-import {Switch, Tabs} from '@gravity-ui/uikit';
+import {Loader, Switch, Tabs} from '@gravity-ui/uikit';
 
 //@ts-ignore
 import TopQueries from './TopQueries/TopQueries';
@@ -53,9 +53,9 @@ function Diagnostics(props: DiagnosticsProps) {
         currentSchema: currentItem = {},
         autorefresh,
     } = useSelector((state: any) => state.schema);
-    const {
-        diagnosticsTab = GeneralPagesIds.overview,
-    } = useSelector((state: any) => state.tenant);
+    const {diagnosticsTab = GeneralPagesIds.overview, wasLoaded} = useSelector(
+        (state: any) => state.tenant,
+    );
 
     const location = useLocation();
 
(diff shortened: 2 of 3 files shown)


src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -53,9 +53,9 @@ function Diagnostics(props: DiagnosticsProps) {
         currentSchema: currentItem = {},
         autorefresh,
     } = useSelector((state: any) => state.schema);
-    const {
-        diagnosticsTab = GeneralPagesIds.overview,
-    } = useSelector((state: any) => state.tenant);
+    const {diagnosticsTab = GeneralPagesIds.overview} = useSelector((state: any) => state.tenant);
+
+    const wasLoaded = useSelector((state: any) => state.tenant.wasLoaded);
```
ariser, line 58: you can join these two selectors

src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -83,7 +83,11 @@ function Diagnostics(props: DiagnosticsProps) {
             return diagnosticsTab;
         } else {
             const newPage = pages[0].id;
-            forwardToDiagnosticTab(newPage);
+
+            if (wasLoaded) {
+                forwardToDiagnosticTab(newPage);
+            }
+
             return newPage;
```
ariser, line 91: this kinda breaks the logic flow
wouldn't it make more sense to check this at the start of this function? And return, say, `undefined` in case `wasLoaded === false`

src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -185,6 +189,13 @@ function Diagnostics(props: DiagnosticsProps) {
         );
     };
 
+    // Loader prevents incorrect loading of tabs
+    // After tabs are initially loaded it is no longer needed
+    // Thus there is no also "loading" check as in other parts of the project
```
ariser, line 194: kudos to you for the explanatory comment

src/containers/Tenant/Diagnostics/Diagnostics.scss
```diff
@@ -1,3 +1,9 @@
+.yc-loader {
+    flex-grow: 1;
+    justify-content: center;
+    align-items: center;
+}
```
ariser, line 5: now here is dangerous stuff
if you need to tweak styles for an element, give it its own classname
or at least nest it in the component selector (own classname still wins)
this one alters the styles for all loaders, possibly on other pages as well, since all styles are bundled

src/containers/Tenant/Diagnostics/Diagnostics.scss
```diff
@@ -1,3 +1,9 @@
+.yc-loader {
+    flex-grow: 1;
+    justify-content: center;
+    align-items: center;
```
ariser, line 4: this works by accident
these properties only work because `.yc-loader` happen to have `display: flex` from the library styles
if it ever changes, this will break

Also I have a feeling that you used these in assumption that they would position `.yc-loader`, whereas in reality they control its content. Sorry if I'm wrong :)

so the simple way is to add `display: flex` to these styles (keeping in mind the other comment about selectors)
but maybe you want to revise the way this loader is positioned
either way is fine by me

ariser (changes requested):

ariser (approved):

PR #202, opened 2022-10-20
feat(Diagnostics): add consumers tab for topics

artemmufazalov:


src/containers/Tenant/Diagnostics/Consumers/Consumers.scss
```diff
@@ -0,0 +1,8 @@
+@import '@gravity-ui/uikit/styles/mixins.scss';
```
ariser, line 1: what is it for?

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,89 @@
...
+                settings={DEFAULT_TABLE_SETTINGS}
+                columns={columns}
+                data={consumersToRender}
+                emptyDataMessage={'No consumers match the current search'}
+            />
+        </div>
+    );
+};
+
+export default Consumers;
```
ariser, line 95: I'd suggest named exports, and `index.ts` for cleaner imports

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,89 @@
...
+            },
+        },
+    ];
+
+    if (consumers.length === 0) {
+        return <div>This topic has no consumers</div>;
+    }
+
+    return (
+        <div>
```
ariser, line 78: the root element should have the block class name

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    const columns: Column<any>[] = [
+        {
+            name: 'name',
+            header: 'Consumers',
+            width: 200,
+            render: ({value}) => {
+                return <span>{value as string}</span>;
+            },
```
ariser, line 69: doesn't it work with the default render?

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+interface ConsumersProps {
+    path: string;
+}
+
+const Consumers = ({path}: ConsumersProps) => {
+    const dispath = useDispatch();
+
+    const fetchData = () => {
+        dispath(getDescribe({path: path}));
```
ariser, line 30: shorthand?

ariser (changes requested):

src/components/Search/Search.tsx
```diff
@@ -1,25 +1,18 @@
 import React from 'react';
```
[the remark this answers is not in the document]
artemmufazalov: For me the key reason for separating search was debounce logic which should be reused. As for the TextInput, it lacks debounce.

As for this HOC, for me it seems to be useless, as it could be replaced with single useState.

So, I think, new Search should be introduced anyway, even if we leave SettingsSearch untouched.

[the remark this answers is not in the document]
ariser: DIscussed this on a call and decided to keep a new search component, but base it on [StorageFilter](https://github.com/ydb-platform/ydb-embedded-ui/tree/main/src/containers/Storage/StorageFilter) instead, which essentially is the same

src/containers/Tenant/Diagnostics/Consumers/Consumers.scss
artemmufazalov: Deleted

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
artemmufazalov: Fixed

src/containers/Storage/Storage.js
```diff
@@ -253,8 +236,12 @@ class Storage extends React.Component {
         return (
             <div className={b('controls')}>
                 <div className={b('search')}>
-                    <StorageFilter
-                        placeholder={storageType === StorageTypes.groups ? 'Group ID, Pool name' : 'Node ID, FQDN'}
+                    <Search
```
artemmufazalov, line 239: Here is a difference in this file, in other places it is a prettier fixes

src/components/Search/Search.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    return (
+        <TextInput
+            controlRef={(node) => {
+                if (typeof inputRef === 'function') {
+                    inputRef(node as HTMLInputElement);
+                } else if (inputRef) {
+                    (inputRef.current as HTMLInputElement) = node as HTMLInputElement;
+                }
+            }}
```
ariser, line 53: i'm not sure if this is useful for this component

artemmufazalov: It was taken from SettingSearch, since it remains as it is, this functionality here is no longer needed

ariser (dismissed):

src/components/Search/Search.tsx
artemmufazalov: Deleted

src/components/Search/Search.tsx
```diff
@@ -1,28 +1,29 @@
...
-interface StorageFilterProps {
-    className?: string;
+interface SearchProps {
+    onChange: (value: string) => void;
     value?: string;
-    placeholder?: string;
-    onChange?: (value: string) => void;
+    className?: string;
     debounce?: number;
+    inputRef?: React.Ref<HTMLInputElement>;
```
ariser, line 10: this prop is unused now

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    useAutofetcher(fetchData, [path]);
+
+    const data = useSelector((state: any) => state.describe.data);
+    const consumersNames: string[] =
+        data[path]?.PathDescription?.PersQueueGroup.PQTabletConfig.ReadRules || [];
+
+    const consumers: Consumer[] = consumersNames.map((name: string) => ({name}));
+
+    const [consumersToRender, setConsumersToRender] = useState<Consumer[]>(consumers);
```
ariser, line 41: I want to comment on a conceptual issue here

looks like `data`, which is a response from backend, is transformed into some shape for the ui component
in this case `data` is not covered with types, and the final shape is described in this very file

but here is a typing that covers the `data` variable:
https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/types/api/schema.ts#L419
it can and should be used here to restrict the selector return value

Now regarding the `Consumer` interface. Forgive me my verboseness while I'm trying to explain in it a way that makes sense.

it is a dedicated interface, and not just a `string[]`, because in the future there will be more columns in the table — that's good
but when there is more columns, this type extends. The transformation becomes more complex, and most likely needs to be moved to a selector. The selector then needs to have a restricted return value type, which would be this `Consumer` interface. Therefore it should be placed in the folder with store types:
https://github.com/ydb-platform/ydb-embedded-ui/tree/main/src/types/store

all this can as well be done now, both the transformation and the typing

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    useEffect(() => {
+        setConsumersToRender(consumers);
+    }, [consumersNames]);
```
ariser, line 45: better memoize `consumerNames` and use it as a dependency here
but if the transformation logic moves to a selector, this becomes even cleaner

ariser (changes requested):

src/store/reducers/describe.ts
```diff
@@ -13,7 +17,10 @@ const describe = function z(state = {loading: false, wasLoaded: false, data: {}}
         }
         case FETCH_DESCRIBE.SUCCESS: {
             const newData = JSON.parse(JSON.stringify(state.data));
-            newData[action.data.Path] = action.data;
+
+            // In  case of successful fetch data should always has some Path, so we could cast type here
```
ariser, line 24: should _have_

What exactly enforces this statement? Looks like a hopeful assumption :)

src/store/reducers/describe.ts
```diff
@@ -42,6 +45,18 @@ const describe = function z(
     }
 };
 
+// Consumers selectors
+const selectConsumersNames = (state: any, path: string): string[] =>
+    state.describe.data[path]?.PathDescription?.PersQueueGroup?.PQTabletConfig?.ReadRules || [];
```
ariser, line 50: same as in the other PR, the base selectors should be reference stable, a new array will cause each dependent selector to recalculate each time

src/store/reducers/describe.ts
```diff
@@ -42,6 +45,18 @@ const describe = function z(
...
+
+export const selectConsumers = createSelector(selectConsumersNames, (names) => {
+    const consumers = names.map((name) => {
+        return {name};
+    });
+
+    return consumers as IConsumer[];
```
ariser, line 57: why cast?

ariser (changes requested):

src/store/reducers/describe.ts
artemmufazalov: It was my way to determine function return type)

As I now understand, it’s could be done differently

ariser (approved):

PR #206, opened 2022-10-25
fix(QueryResultTable): fix full screen table preview

artemmufazalov:
Currently table preview displays only one row. 

This might be due to the default table property is set to `"uniform"`, [so the size of the table is calculated beforehand to fit the viewport](https://github.com/caseywebdev/react-list#type-one-of-simple-variable-or-uniform-defaults-to-simple). Here it probably takes the size of the container and displays only the minimal number of items (1). So the default QueryResultTable `dynamicRenderType` was set to `"variable"` as it was in [QueryEditor](https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/containers/Tenant/QueryEditor/QueryEditor.js).

### DIFF
--- a/src/containers/Tenant/Preview/Preview.scss
+++ b/src/containers/Tenant/Preview/Preview.scss
@@ -52,7 +52,13 @@
     &__result {
         overflow: auto;
 
+        // This fixes last row display for ordinary preview (not fullscreen)
         height: calc(100% - 40px);
         padding: 0 10px;
+
+        // Fix white space footer block for fullscreen preview
+        .kv-fullscreen & {
+            height: 100%;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


artemmufazalov:
The last commit fixes this table display bug

Screen Shot 2022-10-25 at 12 09 30

src/containers/Tenant/Preview/Preview.scss
```diff
@@ -52,7 +52,16 @@
...
     }
 }
+
+// Fix white space footer block for fullscreen preview
+
+.kv-fullscreen {
+    .kv-preview__result {
+        height: 100%;
+    }
+}
```
ariser, line 67: you can write it right next to the not-fullscreen property
`.kv-fullscreen & { ... }`
would be more isolated and easier to maintain

but it's ok either way
resolve this comment or update the pr, and I'll approve

ariser (approved):

PR #208, opened 2022-10-26
fix: add checks for fetch failure with no errors

artemmufazalov:
It turned out, that there could be fetch failure with no errors, and it breaks on attempt to read isCancelled property of undefined error.

Screen Shot 2022-10-26 at 15 44 57

### DIFF
--- a/src/store/reducers/schema.js
+++ b/src/store/reducers/schema.js
@@ -18,7 +18,7 @@ export const initialState = {
     showPreview: false,
 };
 
-const schema = function z(state = initialState, action) {
+const schema = (state = initialState, action) => {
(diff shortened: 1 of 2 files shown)


src/store/reducers/schema.js
```diff
@@ -44,7 +44,7 @@ const schema = function z(state = initialState, action) {
             };
         }
         case FETCH_SCHEMA.FAILURE: {
-            if (action.error.isCancelled) {
+            if (action.error && action.error.isCancelled) {
```
ariser, line 47: you can simply `action.error?.isCancelled`

PR #211, opened 2022-11-01
perf(SchemaTree): batch preloaded data dispatch

ariser:
This fixes a bug with the schema tree.
Expanding a node with a large amount of children caused a visual bug, when only a few first children displayed.
The reason is that children data was saved in the store using a separate dispatch for each child node. In case of hundreds of children this was causing performance issues.
This change alters this behaviour. Child nodes data is now collected and dispatched once in a batch.

### DIFF
--- a/src/store/reducers/schema.js
+++ b/src/store/reducers/schema.js
@@ -2,7 +2,7 @@ import {createRequestActionTypes, createApiRequest} from '../utils';
 import '../../services/api';
 
 const FETCH_SCHEMA = createRequestActionTypes('schema', 'FETCH_SCHEMA');
-const PRELOAD_SCHEMA = 'schema/PRELOAD_SCHEMA';
+const PRELOAD_SCHEMAS = 'schema/PRELOAD_SCHEMAS';
 const SET_SCHEMA = 'schema/SET_SCHEMA';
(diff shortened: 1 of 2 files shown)


artemmufazalov (approved):

src/store/reducers/schema.js
```diff
@@ -134,10 +130,9 @@ export function setShowPreview(value) {
 }
 
 // only stores the passed data if the path doesn't exist yet
-export function preloadSchema(path, data) {
+export function preloadSchemas(data) {
     return {
         type: PRELOAD_SCHEMA,
```
artemmufazalov, line 135: Would it be better to rename action type as well?

artemmufazalov (changes requested):

src/store/reducers/schema.js
```diff
@@ -134,10 +130,9 @@ export function setShowPreview(value) {
 }
 
 // only stores the passed data if the path doesn't exist yet
```
artemmufazalov, line 132: This comment does not correspond to the current realisation

artemmufazalov (changes requested):

artemmufazalov (dismissed):

artemmufazalov (approved):

PR #214, opened 2022-11-08
build: remove renaming static dir to resources

ariser:
Changing the overridden build path to the default `static`. This improves maintainability by sticking to the default naming, which is used in default configs

Backend patch is required in order for this to work, and is on its way


prepare-build.sh
```diff
@@ -1,15 +0,0 @@
-#!/bin/bash
-
-set -e
-
-cd build
-mv static resources
-mv favicon.png resources/
```
ariser, line 7: I moved the favicon to the `static` folder initially, so no transformations are required after build

prepare-build.sh
```diff
@@ -1,15 +0,0 @@
...
-mv media/* resources/media
-rm -r media
```
ariser, line 9: these as well are changes from #186 to fix images being broken; not required with proper paths

prepare-build.sh
```diff
@@ -1,15 +0,0 @@
...
-
-if [ "$(uname)" = "Darwin" ]; then # MacOSX case
-    sed -i '' -e 's/static\///g; s/favicon\.png/resources\/favicon\.png/' index.html
-else # Linux case
-    sed -i -e 's/static\///g; s/favicon\.png/resources\/favicon\.png/' index.html
-fi
```
ariser, line 15: these are two replaces
1. to transform paths like `/resources/static/anything.css/` into `/resource/anything.css` (again, `publicPath` only sets a prefix to all paths, but the target dir was still `static`, hence incorrect paths)
2. to adjust favicon path after moving it a few lines above

2 is no longer needed, since the favicon is initially in the right place
1 is no longer needed, since all paths are correctly created during the bulid

there was also a bug: 1 replaced paths in index.html, but not in styles or scripts, while there also were paths that required similar change

config-overrides.js
```diff
@@ -14,17 +14,6 @@ module.exports = {
...
-                test: /\.svg$/,
-                include: path.resolve(srcRoot, 'assets/illustrations'),
-                loader: 'file-loader',
-                options: {
-                    // default is 'static/media/...', but the embedded version static is served from 'resources/media/...',
-                    // and the 'resources' substring is handled by the publicPath below
-                    name: 'media/[name].[hash:8].[ext]',
-                },
-            });
-            config.output.publicPath = 'resources/';
```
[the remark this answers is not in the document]
ariser: An example of more correct paths.

Imaging we have two pages:
- `/monitoring/tenants`
- `/monitoring/tablets/<tablet_id>`

imagine a file `script.js` loaded on both pages 

in the old version it would load twice, from these urls:
- `/monitoring/resources/js/script.js`
- `/monitoring/tablets/resources/js/script.js`

in the new version it would load once from this url:
- `/monitoring/static/js/script.js`

config-overrides.js
```diff
@@ -14,17 +14,6 @@ module.exports = {
...
             config.output.path = path.resolve(__dirname, 'build/');
```
ariser, line 17: to myself: this could be a redundant override, as `build` is already set as the default folder in `create-react-app`. Maybe consider removing it.

Raubzeug (approved):

config-overrides.js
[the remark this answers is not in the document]
ariser: Correction: discussions revealed that relative paths are crucial for the embedded version.
It can be hosted on a hub, with a proxy backend being at the start of URL, e.g. `/ydb-vla02-dev.yandex.net/monitoring/<...>`.
In this case backend would still receive `filename` as a string starting with `monitoring`, but if in UI paths were absolute, then this request would go to the hub backend, and not to the proxy backend, and will result in 404.

package.json
```diff
@@ -42,7 +42,7 @@
     "start": "react-app-rewired start",
     "dev": "DISABLE_ESLINT_PLUGIN=true TSC_COMPILE_ON_ERROR=true REACT_APP_BACKEND=http://localhost:8765 npm start",
     "build": "react-app-rewired build",
-    "build:embedded": "rm -rf build && PUBLIC_URL=/monitoring/ REACT_APP_BACKEND=http://localhost:8765 npm run build",
+    "build:embedded": "rm -rf build && PUBLIC_URL=. REACT_APP_BACKEND=http://localhost:8765 npm run build",
```
ariser, line 45: I really want to add a comment to this setting, but json file wouldn't allow this. Will find a way to clarify why is it set up this way.

Raubzeug (approved):

Raubzeug (approved):

config-overrides.js
ariser: probably not in this PR

ariser:
Plan for merging:
1. Test this build config with the new backend version before merge
2. Merge as is
3. Release a pre-major version and ask a responsible manager to test the build process
4. Release the major version with release notes describing the difference in paths

PR #215, opened 2022-11-09
Merge CdcStream and PersQueueGroup paths

artemmufazalov:
In this PR `PersQueueGroup` inside `CdcStreams` was merged in `Diagnostics` tab.

To do so functionality for batched requests was added for `createApiRequest` method and  in `describe` and `schema` reducers. `Consumers` tab was enabled to fetch data by child path, while `Overview` and `Describe` tabs to fetch data for both parent and children.


src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -17,18 +19,32 @@ const b = block('ydb-consumers');
...
-export const Consumers = ({path}: ConsumersProps) => {
+export const Consumers = ({path, pathType, schemaNestedChildrenPaths}: ConsumersProps) => {
     const dispath = useDispatch();
 
+    const isCdcStream = pathType === EPathType.EPathTypeCdcStream;
+
+    const targetPath = useMemo(
+        () =>
+            // Since every CDC Stream have only one nested PersQueueGroup, we select only first item
+            (isCdcStream ? schemaNestedChildrenPaths && schemaNestedChildrenPaths[0] : path) ||
```
ariser: this component looks as it is not sure how to behave

either it should know that it works with CDCStreams and PQGroups, and accept a path, selecting required data from store by itself
or it should accept a final path that it would use to make an api call

but right now it requires a selector outside, having at the same time a hardcoded logic for specifically cdcstream inside

let it be one way or another, not both

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -17,18 +19,32 @@ const b = block('ydb-consumers');
...
+        [path, pathType, schemaNestedChildrenPaths],
+    );
+
     const fetchData = () => {
-        dispath(getDescribe({path}));
+        dispath(getDescribe({path: targetPath}));
     };
 
-    useAutofetcher(fetchData, [path]);
+    useAutofetcher(fetchData, [path, pathType, schemaNestedChildrenPaths]);
```
ariser: fetchData does not depend from pathType and schemaNestedChildrenPaths, why are they here?

src/store/reducers/schema.js
```diff
@@ -145,4 +148,29 @@ export function resetLoadingState() {
...
+};
+
+const selectSchemaPathType = (state, path) =>
+    state.schema.data[path]?.PathDescription?.Self?.PathType;
+
+export const selectSchemaChildrenPaths = createSelector(
+    [
+        (_, path) => path,
+        (state, path) => selectSchemaPathType(state, path),
+        (state, path) => selectSchemaChildren(state, path),
```
ariser: why arrows here?

src/store/reducers/schema.js
```diff
@@ -1,5 +1,8 @@
+import {createSelector} from 'reselect';
+
 import {createRequestActionTypes, createApiRequest} from '../utils';
 import '../../services/api';
+import {nestedPaths} from '../../containers/Tenant/utils/schema';
```
ariser: bad import
reducers shouldn't rely on components logic

src/store/reducers/schema.js
```diff
@@ -145,4 +148,29 @@ export function resetLoadingState() {
...
+    [
+        (_, path) => path,
+        (state, path) => selectSchemaPathType(state, path),
+        (state, path) => selectSchemaChildren(state, path),
+    ],
+    (path, pathType, children) => {
+        const nestedChildrenTypes = nestedPaths[pathType];
+        if (nestedChildrenTypes && children) {
+            return children
+                .filter((child) => nestedChildrenTypes.includes(child.PathType))
```
ariser: doesn't this repeat that `checkIfPathNested` function logic?

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -0,0 +1,86 @@
...
+        wasLoaded,
+    } = useSelector((state: any) => state.describe);
+    const {autorefresh, currentSchemaPath} = useSelector((state: any) => state.schema);
+
+    const fetchData = () => {
+        const path = currentSchemaPath || tenant;
+        dispatch(getDescribe({path}));
+    };
+
+    useAutofetcher(fetchData, [], autorefresh);
```
ariser: why empty deps array?

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    useEffect(() => {
+        fetchData();
+    }, [currentSchemaPath]);
```
ariser: `useAutofetcher` does this for you

src/store/reducers/describe.ts
```diff
@@ -2,54 +2,100 @@ import {createSelector, Selector} from 'reselect';
...
         }
         case FETCH_DESCRIBE.FAILURE: {
+            if (action.error?.isCancelled) {
+                return state;
+            }
+
             return {
                 ...state,
                 error: action.error,
+                wasLoaded: true,
```
ariser, line 98: why?

src/store/reducers/describe.ts
```diff
@@ -2,54 +2,100 @@ import {createSelector, Selector} from 'reselect';
...
+
+            let newData: IDescribeData = {},
+                currentDescribe: any = {},
+                wasLoaded = true;
+
+            if (Array.isArray(data)) {
                 newData = JSON.parse(JSON.stringify(state.data));
-                newData[action.data.Path] = action.data;
+
+                data.forEach((item) => {
```
ariser, line 57: all this looks like it belongs to datahandler of the api call

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -16,35 +19,60 @@ const b = cn('kv-describe');
...
+            if (schemaNestedChildrenPaths && schemaNestedChildrenPaths.length > 0) {
+                const paths = [path, ...schemaNestedChildrenPaths];
+                dispatch(
+                    getDescribe({path: paths}, {resetLoadingState, currentDescribePath: path}),
+                );
+            } else {
+                dispatch(getDescribe({path}, {resetLoadingState, currentDescribePath: path}));
+            }
+        } else {
+            dispatch(getDescribe({path}, {resetLoadingState, currentDescribePath: path}));
```
ariser, line 56: aren't these two exactly the same?

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -16,35 +19,60 @@ const b = cn('kv-describe');
...
+        }
+    };
 
-    useEffect(() => {
-        fetchData();
-    }, [currentSchemaPath]);
+    useAutofetcher(
+        (isBackground) => fetchData(!isBackground),
+        [tenant, currentSchemaPath, schemaWasLoaded, schemaNestedChildrenPaths],
+        autorefresh,
```
ariser, line 63: do you have eslint enabled in vscode?
it should warn you about breaking the rules of hooks, which this construction violates. Please follow them

currently linter will warn you about omitting `fetchData` as a dependency, but adding it into the list would make no sense as it is not memoized
but why do you use arrow here and not just `useAutofetcher(fetchData, ...`? This way linter would be more helpful

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -60,7 +88,10 @@ const Describe = ({tenant}: IDescribeProps) => {
         );
     };
 
-    if (loading && !wasLoaded) {
+    // Schema loads children paths.
+    // To prevent rendering describe without children where children are needed,
+    // schema loading state is checked here
+    if (!wasLoaded || (hasNested && schemaLoading && !schemaWasLoaded)) {
```
ariser, line 88: why don't you check describe `loading` anymore?

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -127,7 +143,9 @@ function Overview(props: OverviewProps) {
             [EPathType.EPathTypeExtSubDomain]: undefined,
             [EPathType.EPathTypeColumnStore]: undefined,
             [EPathType.EPathTypeColumnTable]: undefined,
-            [EPathType.EPathTypeCdcStream]: () => <CDCStreamInfo data={schemaData} />,
+            [EPathType.EPathTypeCdcStream]: () => (
+                <CDCStreamInfo data={schemaData} nestedChildren={currentSchemaNestedChildren} />
```
ariser, line 147: I'm really concerned about these “nested children” props and functions everywhere
I can understand their purpose, but I'm not sure about this approach, and I really don't like the naming
this is just to let you know that one of the first comments about “nested something” really goes for all these nested somethings

src/store/reducers/schema.ts
```diff
@@ -39,19 +40,38 @@ const schema: Reducer<ISchemaState, ISchemaAction> = (state = initialState, acti
         case FETCH_SCHEMA.SUCCESS: {
             const newData = JSON.parse(JSON.stringify(state.data));
 
-            if (action.data.Path) {
-                newData[action.data.Path] = action.data;
+            let currentSchemaPath, currentSchema, currentSchemaNestedChildren;
+
+            if (Array.isArray(action.data) && action.data.length > 0) {
```
ariser, line 45: again, heavy calculations, looks like it should be in `dataHandler`

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -26,23 +26,35 @@ interface ConsumersProps {
...
+            isCdcStream ? schemaNestedChildrenPaths && schemaNestedChildrenPaths[0] : path,
         [path, pathType, schemaNestedChildrenPaths],
     );
 
-    const fetchData = () => {
-        dispath(getDescribe({path: targetPath}));
+    const fetchData = (resetLoadingState: boolean) => {
+        // For PersQueueTopic targetPath is defined
+        // For CDC Stream path become defined when schema is loaded
+        // as it loads no its own data, but data of the child
```
ariser, line 44: 'not'

ariser (changes requested):
I'd really appreciate if you made this PR a few separate PRs, two or more. I'm concerned about the changes in reducers, I'd like to review and discuss the new arrays functionality separately. And another thing is this “nested children” stuff.

Having everything in one place makes it harder.

It should be easier for you (than me) to determine which parts of this PR can be made a separate one. Can you do this?

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
artemmufazalov: It seems like I broke some of the commits with fixups. Or, the opposite, I fixed my mistakes by next commits, but not fixups. And it spoils reviewing by commits. I will try to do it more carefully further

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
artemmufazalov: I tried to get rid of this kind of loading check.

My idea is in separating two flags, `loading` and `wasLoaded`, for different purposes. First is for loading data in a background mode, second is for the initial fetch (or fetch on path change), where loader should be shown. And wasLoaded is set only when specific argument is passed to fetch function. 

This way we could do both background fetches or fetches from different components (SchemaTree, Tenant, etc.) without loader reset.

As for the `loading`, I don't know why it is needed. Maybe we can even delete it. But I'm not sure, maybe there was some other logic, that depends on such flags

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
artemmufazalov: I haven't. Thanks for noticing it, now I set up this extension as I should before.

artemmufazalov:
This PR was split with significant changes to other PRs:
1. refactor(Describe): rewrite in TS and hooks, fix loader ([#220](https://github.com/ydb-platform/ydb-embedded-ui/pull/220))
2. fix: optimise requests to json/viewer/describe ([#223](https://github.com/ydb-platform/ydb-embedded-ui/pull/223))
3. refactor(schema): rewrite reducer in TS ([#224](https://github.com/ydb-platform/ydb-embedded-ui/pull/224))
4. fix(Consumers): add loader([#226](https://github.com/ydb-platform/ydb-embedded-ui/pull/226))
5. refactor: apply prettier fixes([#230](https://github.com/ydb-platform/ydb-embedded-ui/pull/230))
6. feat: merge PQGroup and CDCStream ([#231](https://github.com/ydb-platform/ydb-embedded-ui/pull/231))

PR #218, opened 2022-11-14
fix: fix app crash on ColumnTable path type

artemmufazalov:
App crashes on ColumnTable type selection in SchemaTree. The problem is in the functions `prepareOlapTableSchema` and `prepareOlapTableGeneral` that try to take the properties of undefined object or take properties that do not exist.

To fix it I set default value for incoming `tableData`, also I took typings for ColumnTable and ColumnStore from backend for more careful properties check.

There is still some issue with the loading reset for the unloaded data, although it is not as critical as the current issue.

### DIFF
--- a/src/containers/Tenant/Diagnostics/Overview/Overview.tsx
+++ b/src/containers/Tenant/Diagnostics/Overview/Overview.tsx
@@ -12,7 +12,7 @@ import {
     PersQueueGroupInfo,
 } from '../../../../components/InfoViewer/schemaInfo';
 
-import {EPathType} from '../../../../types/api/schema';
+import {EPathType, TColumnTableDescription} from '../../../../types/api/schema';
 import {isColumnEntityType, isTableType} from '../../utils/schema';
 //@ts-ignore
 import {getSchema, resetLoadingState} from '../../../../store/reducers/schema';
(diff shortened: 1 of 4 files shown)


src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -57,19 +62,27 @@ const initialTenantCommonInfoState = {
...
 };
 
-function prepareOlapTableSchema(tableSchema: any) {
-    const {Name, Schema = {}} = tableSchema;
-    const {Columns, KeyColumnNames} = Schema;
-    const KeyColumnIds = KeyColumnNames?.map((name: string) => {
-        const column = Columns?.find((el: any) => el.Name === name);
-        return column.Id;
-    });
+// As result of checks isTableType && isColumnEntityType the incoming type will be TColumnTableDescription
```
Raubzeug, line 65: It seems for me that this comment may be a little bit confusing. You just declare that you need a `TColumnTableDescription` argument for this function and that's it! It doesn't really matter how you check it.

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -25,7 +25,8 @@ import {useAutofetcher} from '../../../../utils/hooks';
 
 import './Overview.scss';
 
-function prepareOlapTableGeneral(tableData: any, olapStats?: any[]) {
+// As result of checks isTableType && isColumnEntityType the incoming type will be TColumnTableDescription
```
Raubzeug, line 28: see my comment to `prepareOlapTableSchema`.

Raubzeug (approved):

Raubzeug (approved):

PR #220, opened 2022-11-15
refactor(Describe): rewrite in TS and hooks, fix loader

artemmufazalov:


src/store/reducers/describe.ts
```diff
@@ -2,54 +2,81 @@ import {createSelector, Selector} from 'reselect';
...
 
-const describe: Reducer<
-    IDescribeState,
-    ApiRequestAction<typeof FETCH_DESCRIBE, TEvDescribeSchemeResult, IResponseError>
-> = (state = initialState, action) => {
+const describe: Reducer<IDescribeState, IDescribeAction> = (state = initialState, action) => {
     switch (action.type) {
         case FETCH_DESCRIBE.REQUEST: {
+            // Reset loading helps to keep correct loader render on tabs or path change
+            const resetLoading = action.additionalParams?.resetLoadingState;
```
Raubzeug, line 28: Let's share responsibility between different actions without additional params, which brings quite a complicated logic.
`REQUEST`, `RESET STATE`, `SET CURRENT DESCRIBE PATH` and so on.

src/store/reducers/describe.ts
```diff
@@ -59,26 +86,28 @@ const describe: Reducer<
...
-const selectConsumersNames: Selector<IDescribeRootStateSlice, string[] | undefined, [string]> = (
-    state,
-    path,
-) => state.describe.data[path]?.PathDescription?.PersQueueGroup?.PQTabletConfig?.ReadRules;
-
-export const selectConsumers: Selector<IDescribeRootStateSlice, IConsumer[]> = createSelector(
-    selectConsumersNames,
-    (names = []) => {
+const selectConsumersNames = (state: IDescribeRootStateSlice, path: string | undefined) =>
+    path
```
Raubzeug, line 90: You may skip this check because `someArray[undefined] = undefined`.

src/store/reducers/describe.ts
```diff
@@ -59,26 +86,28 @@ const describe: Reducer<
...
+        ? state.describe.data[path]?.PathDescription?.PersQueueGroup?.PQTabletConfig?.ReadRules
+        : undefined;
+
+export const selectConsumers: Selector<IDescribeRootStateSlice, IConsumer[], [string | undefined]> =
+    createSelector(selectConsumersNames, (names = []) => {
         const consumers = names.map((name) => {
```
Raubzeug, line 96: you may use more concise way: `names.map((name) => ({name})`

Raubzeug (changes requested):

src/store/reducers/describe.ts
artemmufazalov: It was done to satisfy TypeScript

src/containers/Tenant/Diagnostics/Describe/Describe.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+            dispatch(setCurrentDescribePath(path));
+            dispatch(getDescribe({path}));
+        },
+        [currentSchemaPath, tenant, dispatch],
+    );
+
+    useAutofetcher(fetchData, [fetchData], autorefresh);
+
+    const renderDescribeJson = () => {
```
Raubzeug, line 53: It's overhead to define this function, it only returns component without any additional logic.

src/store/reducers/describe.ts
```diff
@@ -28,21 +27,34 @@ const describe: Reducer<
...
-            let newData;
+            const data = action.data;
 
-            if (action.data.Path) {
+            const isCurrentDescribePath = data.Path === state.currentDescribePath;
+
+            // Do not reset loading state if it is not the current describe path
+            // May be needed in case of schema path change while previous fetch has not been finished
+            const wasLoaded = isCurrentDescribePath || state.wasLoaded;
+            const loading = !wasLoaded;
```
Raubzeug, line 37: it seems these values are independent. When you start a request - `loading` is always true. If you want not to render loader in any cases, you should check it in component.
Or you may write a selector like `selectIsLoadingNewSchema`.

src/store/reducers/describe.ts
```diff
@@ -28,21 +27,34 @@ const describe: Reducer<
...
         case FETCH_DESCRIBE.SUCCESS: {
-            let newData;
+            const data = action.data;
 
-            if (action.data.Path) {
+            const isCurrentDescribePath = data.Path === state.currentDescribePath;
+
+            // Do not reset loading state if it is not the current describe path
+            // May be needed in case of schema path change while previous fetch has not been finished
+            const wasLoaded = isCurrentDescribePath || state.wasLoaded;
```
Raubzeug, line 36: may be `const wasLoaded = isCurrentDescribePath && state.wasLoaded;` ?

src/store/reducers/describe.ts
```diff
@@ -28,21 +27,34 @@ const describe: Reducer<
...
             }
 
             return {
                 ...state,
                 data: newData,
-                currentDescribe: action.data,
-                loading: false,
-                wasLoaded: true,
+                currentDescribe: currentDescribe,
+                loading: loading,
```
Raubzeug, line 56: No need to duplicate value, if key name = value name.
`{loading,
wasLoaded}`

src/store/reducers/describe.ts
```diff
@@ -53,27 +65,47 @@ const describe: Reducer<
...
 };
 
+export const setCurrentDescribePath = (path: string) => {
+    return {
+        type: SET_CURRENT_DESCRIBE_PATH,
+        data: path,
+    } as const;
+};
+
+export const resetLoadingState = () => {
```
Raubzeug, line 95: may be `setDataWasNotLoaded` or something like that will be more meaningful?

src/store/reducers/describe.ts
```diff
@@ -53,27 +65,47 @@ const describe: Reducer<
...
+    return {
+        type: RESET_LOADING_STATE,
```
Raubzeug, line 97: more meaningful name?

src/store/reducers/describe.ts
```diff
@@ -2,24 +2,23 @@ import {createSelector, Selector} from 'reselect';
...
+    currentDescribePath: undefined,
 };
 
-const describe: Reducer<
-    IDescribeState,
-    ApiRequestAction<typeof FETCH_DESCRIBE, TEvDescribeSchemeResult, IResponseError>
-> = (state = initialState, action) => {
+const describe: Reducer<IDescribeState, IDescribeAction> = (state = initialState, action) => {
     switch (action.type) {
         case FETCH_DESCRIBE.REQUEST: {
```
Raubzeug, line 23: What do think if we check if path from props is not equal to `currentDescribePath` here? And if it is not we can set `currentDescribePath:path`, `data: undefined` and `wasLoaded: false` ?

Raubzeug (changes requested):

src/store/reducers/describe.ts
```diff
@@ -28,21 +27,34 @@ const describe: Reducer<
             };
         }
         case FETCH_DESCRIBE.SUCCESS: {
-            let newData;
+            const data = action.data;
 
-            if (action.data.Path) {
+            const isCurrentDescribePath = data.Path === state.currentDescribePath;
+
+            // Do not reset loading state if it is not the current describe path
```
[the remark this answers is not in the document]
artemmufazalov: How do we cancel previous requests?

src/store/reducers/describe.ts
artemmufazalov: No, if it is not a current path, we just take the value that is in state

src/store/reducers/describe.ts
artemmufazalov: Changed it to ` const loading = isCurrentDescribePath ? false : state.loading`

src/store/reducers/describe.ts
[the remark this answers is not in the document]
Raubzeug: `AbortController` may help with this

src/store/reducers/describe.ts
```diff
@@ -28,21 +27,34 @@ const describe: Reducer<
             };
         }
         case FETCH_DESCRIBE.SUCCESS: {
-            let newData;
+            const data = action.data;
 
-            if (action.data.Path) {
+            const isCurrentDescribePath = data.Path === state.currentDescribePath;
```
Raubzeug, line 32: May be it will be more clear if we check:
```
if (!isCurrentDescribePath) {
return {...state}
}
```

src/store/reducers/describe.ts
```diff
@@ -53,27 +65,47 @@ const describe: Reducer<
...
+        case SET_CURRENT_DESCRIBE_PATH: {
+            // Current describe should correspond to current path,
+            // so currentDescribePath is set to check it
+            return {
+                ...state,
+                currentDescribePath: action.data,
+            };
+        }
+        case SET_DATA_WAS_NOT_LOADED: {
+            // Reset loading helps to keep correct loader render on tabs or path change
```
Raubzeug, line 77: Not sure this comment is essential.

Raubzeug (changes requested):

src/store/reducers/describe.ts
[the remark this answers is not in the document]
artemmufazalov: Aborting describe requests that do not correspond to current path could break describe requests outside of Describe component.

For example, currently describe is also used in consumers tab.

Also, as we try to merge CDC Streams and PQGroups tabs for CDC Streams, in consumers will be requested describe that does not correspond to current path (current path - CDC Stream, requested describe - PQGroup)

src/store/reducers/describe.ts
artemmufazalov: Thanks, it is much better

Raubzeug (approved):

Raubzeug (approved):

PR #225, opened 2022-11-21
fix(Storage): fix uneven PDisks

artemmufazalov:


Raubzeug (approved):

artemmufazalov:
> Let's try to check maximum quantity of pDisks in the row and calculate pDisk width on depending of this. This will allow to stretch pDisks to the full column width.

As we plan to replace virtual scroll with data partition in backend responses, we won’t be able to get maximum PDisks length. So, we decided to stick to the current solution

PR #226, opened 2022-11-21
fix(Consumers): add loader

artemmufazalov:


src/containers/Tenant/Diagnostics/Consumers/Consumers.scss
```diff
@@ -3,4 +3,11 @@
         width: 200px;
         margin-bottom: 20px;
     }
+    &__loader {
```
Raubzeug, line 6: I see this styles in many places where uikit loader is used. May be it would be reasonable to create an internal Loader component and use it throughout the app?

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -59,8 +77,20 @@ export const Consumers = ({path}: ConsumersProps) => {
...
 
+    if (loading && !wasLoaded) {
+        return (
+            <div className={b('loader')}>
+                <Loader size="m" />
+            </div>
+        );
+    }
+
+    if (!loading && error) {
```
Raubzeug, line 88: case when `!loading && error` should not be real. Please check your reducer to make sure that error is set to undefined when start a new request.

artemmufazalov: Actually, it is set to undefined on request success. Isn't that enough?

Raubzeug: Discussed in PM, this is done to avoid error text blinking when requests are failing in a row when autofetcher is enabled.

Raubzeug (approved):

PR #231, opened 2022-11-29
feat: merge PQGroup and CDCStream

artemmufazalov:
In this PR data for `PersQueueGroup` and `CdcStreams` was merged in `Diagnostics` tabs (Overview, Consumers, Describe). Object of type PersQueueGroup is no longer visible.

To do so functionality for batched requests was added for `createApiRequest` method and  in `describe` and `schema` reducers. `Consumers` tab was enabled to fetch data by child path, while `Overview` and `Describe` tabs to fetch data for both parent and children.

### DIFF
--- a/src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
+++ b/src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
@@ -1,29 +1,35 @@
 import type {TEvDescribeSchemeResult, TCdcStreamDescription} from '../../../types/api/schema';
+import {useTypedSelector} from '../../../utils/hooks';
+import {selectSchemaData} from '../../../store/reducers/schema';
 
-import {formatCdcStreamItem, formatCommonItem} from '../formatters';
+import {formatCdcStreamItem, formatPQGroupItem, formatCommonItem} from '../formatters';
 import {InfoViewer, InfoViewerItem} from '..';
 
 const DISPLAYED_FIELDS: Set<keyof TCdcStreamDescription> = new Set(['Mode', 'Format']);
 
 interface CDCStreamInfoProps {
     data?: TEvDescribeSchemeResult;
+    childrenPaths?: string[];
 }
 
-export const CDCStreamInfo = ({data}: CDCStreamInfoProps) => {
-    if (!data) {
-        return <div className="error">No CDC Stream data</div>;
+export const CDCStreamInfo = ({data, childrenPaths}: CDCStreamInfoProps) => {
+    const pqGroupData = useTypedSelector((state) => selectSchemaData(state, childrenPaths?.[0]));
+
+    if (!data || !pqGroupData) {
+        return <div className="error">No Changefeed data</div>;
     }
 
-    const TableIndex = data.PathDescription?.CdcStreamDescription;
+    const cdcStream = data.PathDescription?.CdcStreamDescription;
+    const pqGroup = pqGroupData?.PathDescription?.PersQueueGroup;
+
     const info: Array<InfoViewerItem> = [];
 
-    info.push(formatCommonItem('PathType', data.PathDescription?.Self?.PathType));
     info.push(formatCommonItem('CreateStep', data.PathDescription?.Self?.CreateStep));
 
     let key: keyof TCdcStreamDescription;
-    for (key in TableIndex) {
+    for (key in cdcStream) {
         if (DISPLAYED_FIELDS.has(key)) {
-            info.push(formatCdcStreamItem(key, TableIndex?.[key]));
+            info.push(formatCdcStreamItem(key, cdcStream?.[key]));
         }
     }
 
--- a/src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
+++ b/src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
@@ -27,5 +27,13 @@ export const CDCStreamInfo = ({data}: CDCStreamInfoProps) => {
         }
     }
 
-    return <>{info.length ? <InfoViewer info={info}></InfoViewer> : <>Empty</>}</>;
+    info.push(formatPQGroupItem('Partitions', pqGroup?.Partitions || []));
+    info.push(
+        formatPQGroupItem(
+            'PQTabletConfig',
+            pqGroup?.PQTabletConfig || {PartitionConfig: {LifetimeSeconds: 0}},
+        ),
+    );
+
+    return <InfoViewer title={'Changefeed'} info={info} />;
 };
--- a/src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
+++ b/src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
@@ -4,7 +4,8 @@ import block from 'bem-cn-lite';
 
 import DataTable, {Column} from '@yandex-cloud/react-data-table';
 
-import { Loader } from '../../../../components/Loader';
+import type {EPathType} from '../../../../types/api/schema';
+import {Loader} from '../../../../components/Loader';
 import {prepareQueryError} from '../../../../utils/query';
 import {DEFAULT_TABLE_SETTINGS} from '../../../../utils/constants';
 import {useAutofetcher, useTypedSelector} from '../../../../utils/hooks';
@@ -15,6 +16,9 @@ import {
     setCurrentDescribePath,
     setDataWasNotLoaded,
 } from '../../../../store/reducers/describe';
+import {selectSchemaMergedChildrenPaths} from '../../../../store/reducers/schema';
+
+import {isCdcStreamEntityType} from '../../utils/schema';
 
 import i18n from './i18n';
 
@@ -24,21 +28,33 @@ const b = block('ydb-consumers');
 
 interface ConsumersProps {
     path: string;
+    type?: EPathType;
 }
 
-export const Consumers = ({path}: ConsumersProps) => {
+export const Consumers = ({path, type}: ConsumersProps) => {
     const dispatch = useDispatch();
+
+    const isCdcStream = isCdcStreamEntityType(type);
+
+    const mergedChildrenPaths = useTypedSelector((state) =>
+        selectSchemaMergedChildrenPaths(state, path, type),
+    );
+
+    const dataPath = isCdcStream ? mergedChildrenPaths?.[0] : path;
 
     const fetchData = useCallback(
         (isBackground: boolean) => {
             if (!isBackground) {
                 dispatch(setDataWasNotLoaded());
             }
-            dispatch(setCurrentDescribePath(path));
-            dispatch(getDescribe({path}));
+
+            if (dataPath) {
+                dispatch(setCurrentDescribePath(dataPath));
+                dispatch(getDescribe({path: dataPath}));
+            }
         },
 
-        [path, dispatch],
+        [dispatch, dataPath],
     );
 
     const {autorefresh} = useTypedSelector((state) => state.schema);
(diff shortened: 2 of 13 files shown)


src/types/store/describe.ts
```diff
@@ -18,9 +18,15 @@ export interface IDescribeState {
     error?: IResponseError;
 }
 
+export interface IDescribeHandledlResponse {
```
ariser, line 21: Handledl

src/store/utils.ts
```diff
@@ -42,7 +42,14 @@ export function createApiRequest<
         });
 
         try {
-            const result = await request;
+            let result;
+
+            if (Array.isArray(request)) {
+                result = await Promise.all(request);
```
ariser, line 48: In case any of the requests fails, the catch statement would receive only the first error. Potentially, this could lead to unevident bugs. E.g. out of two requests the first one fails because of a transport error, whereas the second one answers with HTTP 401. This should result in the auth form displayed, but will only produce an error toast.

src/store/reducers/describe.ts
```diff
@@ -106,10 +105,48 @@ const selectConsumersNames = (state: IDescribeRootStateSlice, path: string | und
...
+        request = path.map((p) => window.api.getDescribe({path: p}));
+    } else {
+        request = window.api.getDescribe({path});
+    }
+
+    return createApiRequest<
+        typeof FETCH_DESCRIBE,
+        TEvDescribeSchemeResult,
+        IDescribeHandledlResponse
+    >({
```
ariser, line 120: these types are inferred, why explicitly set them?

src/store/reducers/describe.ts
```diff
@@ -106,10 +105,48 @@ const selectConsumersNames = (state: IDescribeRootStateSlice, path: string | und
...
+        request,
         actions: FETCH_DESCRIBE,
+        dataHandler: (data) => {
+            let dataPath: string | undefined;
+            let currentDescribe: IDescribeData | TEvDescribeSchemeResult = {};
+            const newData: IDescribeData = {};
+
+            if (Array.isArray(data)) {
```
ariser, line 128: in this case `request` can be both an array and a single promise, so the typeof `data` inferred as `Response | Response[]`, and it's fine

but in case of a single request I want the type of `data` to be `Result` (and vice versa), TS should be smart enough to know. Currently this isn't the case, the type is inferred as `Result | Result[]` even for a single request

src/types/store/describe.ts
```diff
@@ -18,9 +18,15 @@ export interface IDescribeState {
...
+    path: string | undefined;
+    data: IDescribeData | undefined;
+    currentDescribe: TEvDescribeSchemeResult | IDescribeData | undefined;
```
ariser, line 24: complex types are always worse than simple ones
this can be always represented as `IDescribeData`, no need to convert to a plain value for the edge case of a single item

src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
```diff
@@ -1,31 +1,47 @@
...
 }
 
-export const CDCStreamInfo = ({data}: CDCStreamInfoProps) => {
-    if (!data) {
-        return <div className="error">No CDC Stream data</div>;
+export const CDCStreamInfo = ({data, childrenPaths}: CDCStreamInfoProps) => {
+    const pqGroupData = useTypedSelector((state) => selectSchemaData(state, childrenPaths?.[0]));
+
+    if (!data || !pqGroupData) {
+        return <div className="error">No Changefeed data</div>;
```
ariser, line 19: is this the approved wording? Why «Changefeed» is capitalized? Is it a clear term?

src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
```diff
@@ -1,31 +1,47 @@
...
+    info.push(formatPQGroupItem('Partitions', pqGroup?.Partitions || []));
+    info.push(
+        formatPQGroupItem(
+            'PQTabletConfig',
+            pqGroup?.PQTabletConfig || {PartitionConfig: {LifetimeSeconds: 0}},
+        ),
+    );
+
+    return (
+        <>{info.length ? <InfoViewer title={'Changefeed'} info={info}></InfoViewer> : <>Empty</>}</>
```
ariser, line 45: could be `<InfoViewer />`

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -78,14 +80,26 @@ function Overview(props: OverviewProps) {
 
     const loading = schemaLoading || olapStatsLoading;
 
-    useAutofetcher(
-        (isBackground) => {
+    const childrenPaths = useTypedSelector(
+        (state) => selectSchemaChildrenPaths(state, currentSchemaPath),
+        shallowEqual,
```
ariser, line 85: no need for shallow equality, we use reselect
if unwanted rerenders occur, this means there is an issue with how selectors are written

src/store/reducers/schema.ts
```diff
@@ -114,10 +118,45 @@ const schema: Reducer<ISchemaState, ISchemaAction> = (state = initialState, acti
...
-        request: window.api.getSchema({path}),
+export function getSchema({path}: {path: string | string[]}) {
+    let request;
+
+    if (Array.isArray(path)) {
+        request = path.map((p) => window.api.getSchema({path: p}));
+    } else {
+        request = window.api.getSchema({path});
+    }
+    return createApiRequest<typeof FETCH_SCHEMA, TEvDescribeSchemeResult, ISchemaHandledResponse>({
```
ariser, line 129: same with inferred types

src/store/reducers/schema.ts
```diff
@@ -162,6 +201,9 @@ export function resetLoadingState() {
 export const selectSchemaChildren = (state: ISchemaRootStateSlice, path: string | undefined) =>
     path ? state.schema.data[path]?.PathDescription?.Children : undefined;
 
+export const selectSchemaData = (state: ISchemaRootStateSlice, path: string | undefined) =>
+    path ? state.schema.data[path] : undefined;
```
ariser, line 205: same with undefined path, `state.schema.data[path]` would automatically return undefined

ariser (changes requested):

src/store/reducers/describe.ts
ariser: this should be addressed in combination with this comment: https://github.com/ydb-platform/ydb-embedded-ui/pull/231#discussion_r1034701826

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
ariser: This comment is for the future reference.

discussed this on a call; selectors are fine, the issue is unconditionally updating a field in the store, which causes reference update, so reselect can't help. This is an issue because of a circular reference: data fetch depends on children paths, and children paths are updated by this data fetch

shallow equality fixes this, so it is ok to keep it
but a better way would be to either get rid of the circular reference, or at least to move the shallow equality comparator from the component to the selector declaration. There were some issues with the latter previously though. So again, it is ok to keep it as is, if nothing else works.

src/store/reducers/schema.ts
```diff
@@ -153,4 +159,18 @@ export function resetLoadingState() {
     } as const;
 }
 
+export const selectSchemaChildren = (state: ISchemaRootStateSlice, path: string | undefined) =>
+    path ? state.schema.data[path]?.PathDescription?.Children : undefined;
```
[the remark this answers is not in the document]
artemmufazalov: In this case TS throws an error: Type 'undefined' cannot be used as an index type. So I did this

src/store/reducers/schema.ts
```diff
@@ -153,4 +159,18 @@ export function resetLoadingState() {
     } as const;
 }
 
+export const selectSchemaChildren = (state: ISchemaRootStateSlice, path: string | undefined) =>
```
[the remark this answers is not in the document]
artemmufazalov: Yes, fixed, It looks more concise

src/store/reducers/schema.ts
```diff
@@ -153,4 +159,18 @@ export function resetLoadingState() {
...
+    path ? state.schema.data[path]?.PathDescription?.Children : undefined;
+
+export const selectSchemaChildrenPaths: Selector<
+    ISchemaRootStateSlice,
+    string[] | undefined,
+    [string | undefined]
```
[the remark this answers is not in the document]
artemmufazalov: In the Selector generic returnType is the second type (`string[] | undefined` in this case), the third is array of selector params types (in this case `string | undefined` is type of path param)

src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
artemmufazalov: We agreed to name CDCStream Changefeed as it is in docs. Since it is schema object special term, I decided to make it capitalised

src/store/reducers/schema.ts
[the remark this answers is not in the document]
ariser: Ah, I see. Ok :)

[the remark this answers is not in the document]
ariser: Ah, I see. Ok :)

ariser (changes requested):

src/components/InfoViewer/schemaInfo/CDCStreamInfo.tsx
```diff
@@ -41,7 +41,5 @@ export const CDCStreamInfo = ({data, childrenPaths}: CDCStreamInfoProps) => {
         ),
     );
 
-    return (
-        <>{info.length ? <InfoViewer title={'Changefeed'} info={info}></InfoViewer> : <>Empty</>}</>
-    );
+    return <InfoViewer title={'Changefeed'} info={info}></InfoViewer>;
```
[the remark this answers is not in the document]
artemmufazalov: I misunderstood you. The reason, why I removed the check for the empty state here is that it is checked inside InfoViewer

src/containers/Tenant/Diagnostics/Consumers/Consumers.tsx
```diff
@@ -24,29 +28,41 @@ const b = block('ydb-consumers');
...
 interface ConsumersProps {
     path: string;
+    type?: EPathType;
 }
 
-export const Consumers = ({path}: ConsumersProps) => {
+export const Consumers = ({path, type}: ConsumersProps) => {
     const dispatch = useDispatch();
 
+    const isCdcStream = isCdcStreamEntityType(type);
```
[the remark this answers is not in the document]
artemmufazalov: Yes, it is. and it can be replaced with new check. But I decided to left it as it is (with isCDCStream), because it is topics and stream specific tab, and data loading logic is different

src/store/reducers/schema.ts
```diff
@@ -204,14 +204,20 @@ export const selectSchemaChildren = (state: ISchemaRootStateSlice, path: string
...
-    [(_, path: string | undefined) => path, selectSchemaChildren],
-    (path, children) => {
-        return children?.map(({Name}) => path + '/' + Name);
+    [
+        (_, path?: string) => path,
+        (_, _path, isEntityWithMergedImplementation: boolean) => isEntityWithMergedImplementation,
+        selectSchemaChildren,
+    ],
+    (path, isEntityWithMergedImplementation, children) => {
+        return isEntityWithMergedImplementation
```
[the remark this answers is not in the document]
artemmufazalov: I can import check from [schema](https://github.com/ydb-platform/ydb-embedded-ui/blob/06d26def15496f8e2de00d941b39bf6a68382f14/src/containers/Tenant/utils/schema.ts) and pass the type here. But it will be an import from components. How do you think, where should I place schema, to keep imports more clear?

ariser (dismissed):

ariser (approved):

PR #232, opened 2022-11-29
fix(InfoViewer): allow multiline values

ariser:


artemmufazalov:
These changes affect some other components, where InfoViewer is used. In most of the cases it is fine, except the cases, where some objects (not text) is used as value. The examples are ClusterInfo, PDisk info in NodeStructure, VDisk tooltip in node structure. Somewhere different objects are aligned differently, somewhere it looks not as good as before.
#### Before
Before
#### After
After
#### Different align
Align

ariser:
Thanks, good catch.
Changed the alignment to baseline. This fixes the first case with ClusterInfo. However, I consider the case in node structure is fine, the previous alignment with green squares being positioned in the middle of the dots looks weird.

artemmufazalov (approved):

PR #233, opened 2022-12-01
fix(Diagnostics): display DB tabs for nested DBs

artemmufazalov:


src/containers/Tenant/Diagnostics/Compute/Compute.js
```diff
@@ -50,7 +49,17 @@ class Compute extends React.Component {
     }
 
     componentDidUpdate(prevProps) {
-        const {autorefresh, getNodes, tenantName} = this.props;
+        const {autorefresh, getNodes, setDataWasNotLoaded, tenantName} = this.props;
+
+        if (tenantName !== prevProps.tenantName) {
+            setDataWasNotLoaded();
+            getNodes(tenantName);
```
ariser, line 56: now this is ok
but the code is duplicated, maybe separate it into some functions here, as you did in Network?

ariser (changes requested):

src/containers/Tenant/Diagnostics/Compute/Compute.js
```diff
@@ -50,12 +49,25 @@ class Compute extends React.Component {
     }
 
     componentDidUpdate(prevProps) {
-        const {autorefresh, getNodes, tenantName} = this.props;
+        const {autorefresh, getNodes, setDataWasNotLoaded, tenantName} = this.props;
 
-        if (autorefresh && !prevProps.autorefresh) {
-            getNodes(tenantName);
+        const restartAutorefresh = () => {
```
artemmufazalov, line 54: Added function restartAutorefresh here

ariser (approved):

PR #234, opened 2022-12-02
chore: cover nodes api call with types

artemmufazalov:
Add types for request to `/viewer/json/nodes`. Currently this request is part of `getStorage`, but further it will replace getNodes api call (currently `/viewer/json/compute`).

Also storage type file was split to storage, pdisk, vdisk and enums (common enums used in different types), corresponding imports was updated.


src/types/api/enums.ts
```diff
@@ -0,0 +1,7 @@
+export enum EFlag {
```
ariser, line 1: what is `EFlag` and which api handle does it belong to?
separating large the files makes sense, but not at the cost of clarity
the original structure implied describing types per api handle
if you separate them, I suggest make a folder

src/containers/Storage/StorageGroups/StorageGroups.tsx
```diff
@@ -123,8 +129,7 @@ function StorageGroups({data, tableSettings, visibleEntities, nodes, onShowAll}:
             header: tableColumnsNames[TableColumnsIds.Type],
             render: ({value, row}) => (
                 <>
-                    <Label>{value as string || '—'}</Label>
-                    {' '}
+                    <Label>{(value as string) || '—'}</Label>{' '}
```
ariser, line 132: why?

ariser (changes requested):

src/types/api/enums.ts
ariser: >  storage type file was split to storage, pdisk, vdisk and enums (common enums used in different types

is `EFlag` used in responses for api calls other than `viewer/json/storage`?
if so, it is ok to have it in a shred file, but then it lacks a descriptive comment

artemmufazalov: This flag shows system status.

Currently it is in `viewer/json/nodes` and `viewer/json/storage` endpoints. 

If we will add be adding more types, it may appear in:

`/viewer/json/tenantinfo`
`/viewer/json/cluster`
`/viewer/json/tabletinfo`
`/viewer/json/compute`

artemmufazalov: As for the separate VDisk and PDisk types, there are api calls `/viewer/json/pdiskinfo` and `/viewer/json/vdiskinfo`. I'm not sure if types from storage fully correspond to responses from this endpoints (i didn't check this), but I suppose there will be a lot of interconnections. Also Pdisk type is used in `viewer/json/nodes`

src/containers/Storage/StorageGroups/StorageGroups.tsx
artemmufazalov: Fix made by prettier. I didn't paid enough attention, it looked better previously. Will fix it

src/types/api/enums.ts
ariser: ok then
I'd still like to have some sort of distinction between files that describe a certain endpoint, and files that describe shared types. Just for clarity. I'll approve this PR, so if you don't have any ideas right now, we do this some other time. But feel free to update.

ariser (approved):

PR #236, opened 2022-12-05
chore: cover path domain description with types

artemmufazalov:


src/types/api/schema.ts
```diff
@@ -85,6 +85,81 @@ export interface TDirEntry {
...
+interface TProcessingParams {
+    Version?: number;
+    /** uint64 */
+    PlanResolution?: string;
+    /** fixed64 */
+    Coordinators: string[];
+    /** uint64 */
+    TimeCastBucketsPerMediator?: string;
+    /** fixed64 */
+    Mediators: string[];
```
ariser, line 132: `repeated` fields in protobuf are also optional

src/types/api/schema.ts
```diff
@@ -85,6 +85,81 @@ export interface TDirEntry {
     Version?: TPathVersion;
 }
 
+interface TDomainDescription {
```
ariser, line 88: I think some fields from the protobuf are missing. Is it intentional?

ariser (changes requested):

src/types/api/schema.ts
artemmufazalov: I assumed, that in case of empty data we will get empty array, as these fields goes without `optional`. Or it could be a case, when there won't be these fields?

ariser: in protobuf, an empty array and the absence of a field are indistinguishable
they can be omitted

src/types/api/schema.ts
artemmufazalov: Added this missing types

ariser: ok, I think this is fine, but we may want to mark these interfaces as incomplete, like some others currently are marked

ariser (approved):

PR #240, opened 2022-12-07
fix: display object name in overview

artemmufazalov:


src/containers/Tenant/Schema/SchemaInfoViewer/SchemaInfoViewer.js
```diff
@@ -151,36 +157,43 @@ class SchemaInfoViewer extends React.Component {
             );
         }
 
-        if ([tabletMetricsInfo, partitionConfigInfo, tableStatsInfo.flat()].flat().length === 0) {
-            return <div className={b('item')}>Empty</div>;
+        if (
+            [generalTableInfo, tabletMetricsInfo, partitionConfigInfo, tableStatsInfo.flat()].flat()
+                .length === 0
+        ) {
```
ariser, line 163: I don't think this looks better than before. I understand this is what prettier does, but again, seems like we need to tune it. RN a solution might be to make it a variable

src/containers/Tenant/utils/schema.ts
```diff
@@ -41,6 +43,49 @@ export const mapPathTypeToNavigationTreeType = (
...
+
+const databaseTypeToDBName: Record<ETenantType, string | undefined> = {
+    [ETenantType.UnknownTenantType]: 'Database',
+    [ETenantType.Domain]: 'Cluster Root',
+    [ETenantType.Dedicated]: 'Dedicated Database',
+    [ETenantType.Shared]: 'Shared Database',
+    [ETenantType.Serverless]: 'Serverless Database',
+};
+
+export const mapDatabaseTypeToDBName = (type?: ETenantType) => type && databaseTypeToDBName[type];
```
ariser, line 85: what about i18n?

ariser (changes requested):

src/containers/Tenant/utils/schema.ts
artemmufazalov: I don't really like the idea of translating special terms. Even if we decide to translate them, we should discuss correct naming together

ariser: If they have translation in the docs, I believe they should be translated in UI. If not, then discussable. As for coming up with the naming — yes, of course we should decide with the team.

ariser (approved):

PR #241, opened 2022-12-07
refactor(TopShards): migrate reducer to ts

ariser:


src/store/reducers/shardsWorkload.ts
```diff
@@ -1,22 +1,30 @@
...
+    order: string;
+}
+
+function formatSortOrder({columnId, order}: SortOrder) {
     return `${columnId} ${order}`;
 }
 
-function createShardQuery(path, sortOrder, tenantName) {
+function createShardQuery(path: string, sortOrder?: SortOrder[], tenantName?: string) {
     const orderBy = Array.isArray(sortOrder) ?
```
artemmufazalov, line 28: `sortOrder` is of Array type as function requires it. So we can check just `sortOrder ? ... : ""`

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #245, opened 2022-12-08
refactor(ProblemFilter): migrate to ts

artemmufazalov:


src/utils/constants.ts
```diff
@@ -118,6 +118,8 @@ export const COLORS_PRIORITY = {
 export const ALL = 'All';
 export const PROBLEMS = 'With problems';
 
+export type IProblemFilterValues = typeof ALL | typeof PROBLEMS;
```
ariser, line 121: this is `string | string`, i believe what you wanted is `ALL | PROBLEMS`

ariser (changes requested):

src/utils/constants.ts
artemmufazalov: No, it works just fine.The type of `typeof ALL | typeof PROBLEMS` will be `"All" | "With problems"` as expected.

In case of ALL | PROBLEMS there will be an error - 'ALL' (or 'PROBLEMS') refers to a value, but is being used as a type here. Did you mean 'typeof ALL' (or 'PROBLEMS')?

ariser: Yep, you're right

ariser (approved):

PR #249, opened 2022-12-15
feat(Nodes): migrate to viewer/json/nodes

artemmufazalov:


artemmufazalov:
Some changes connected with this PR was merged separately. They include:
1. #244
2. #254
3. #255
4. #264
5. #266 
6. #267

src/containers/Nodes/Nodes.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    const nodes = useTypedSelector(getFilteredPreparedNodesList);
+
+    const useNodesEndpoint = useTypedSelector((state) =>
+        getSettingValue(state, USE_NODES_ENDPOINT_IN_DIAGNOSTICS_KEY),
+    );
+
+    const fetchNodes = useCallback(() => {
+        if (tenantPath && !JSON.parse(useNodesEndpoint)) {
+            dispatch(getComputeNodes(tenantPath));
```
artemmufazalov, line 78: On older cluster versions getNodes doesn't work as intended in Diagnostics. To fix it by default nodes requested from /compute endpoint if tenantPath is present. 

It can be altered by `USE_NODES_ENDPOINT_IN_DIAGNOSTICS` flag. It is made primarily for developers, to be able to easily check whether clusters versions are update and interface is ready to migrate completely. It can be triggered inside Experiments in Settings

src/store/reducers/nodes.ts
```diff
@@ -42,22 +57,25 @@ const nodes: Reducer<INodesState, INodesAction> = (state = initialState, action)
                 loading: false,
             };
         }
-        case CLEAR_NODES: {
+        case RESET_NODES_STATE: {
```
Raubzeug, line 60: data and totalNodes should be persistent when state resets?

Raubzeug (changes requested):

src/containers/Nodes/Nodes.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+import i18n from './i18n';
+
+const b = cn('ydb-nodes');
+
+interface NodesProps {
+    tenantPath?: string;
+    className?: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    additionalNodesInfo: any;
```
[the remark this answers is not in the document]
artemmufazalov: Thanks. This comment helped me to discover that i missed 1 property in `INodesPreparedEntity`, that is used by function getNodeRef

src/containers/Nodes/Nodes.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    const handleProblemFilterChange = (value: string) => {
+        dispatch(changeFilter(value));
+    };
+
+    const handleUptimeFilterChange = (value: string) => {
+        dispatch(setNodesUptimeFilter(value as NodesUptimeFilterValues));
+    };
+
+    const renderControls = () => {
```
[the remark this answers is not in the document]
artemmufazalov: I'm not sure, that it will be more concise. I counted 11 values, that will be passed as props if controls are separate

src/store/reducers/nodes.ts
artemmufazalov: Yes. I will make state reset more explicit, so it will be clear which values are reset

src/containers/Nodes/Nodes.tsx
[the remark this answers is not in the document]
Raubzeug: It may be done with Context api, but I don't insist on it. It's just a suggestion for maybe further decisions.

src/store/reducers/nodes.ts
Raubzeug: I didn't meant to set an explicit list of params. I just noticed that data and totalNodes are not reseted and wanted to clarify that it is correct...

artemmufazalov: I think now it's better as it won't change, if some new values will be added to initial state. For example, default empty object for data, that will reset data on path change. Maybe I can set variables not as constants, but as `InitialState.loading` or `InitialState.searchValue` and etc.

PR #251, opened 2022-12-15
refactor(Tablets): migrate reducer to ts

artemmufazalov:


src/types/store/tablets.ts
```diff
@@ -0,0 +1,42 @@
...
+    IResponseError
+>;
+
+export type ITabletsAction =
+    | ITabletsApiRequestAction
+    | (
+          | ReturnType<typeof clearWasLoadingFlag>
+          | ReturnType<typeof setStateFilter>
+          | ReturnType<typeof setTypeFilter>
+      );
```
ariser, line 38: why group the ReturnTypes?

artemmufazalov: No actual reason. I did this previously, it seems more beautiful for me

PR #258, opened 2022-12-19
fix(InfoViewer): min width for values

ariser:
Without the minimum width values in the overview section shrinked to a one-char column

### DIFF
--- a/src/components/InfoViewer/InfoViewer.scss
+++ b/src/components/InfoViewer/InfoViewer.scss
@@ -54,6 +54,8 @@
     &__value {
         display: flex;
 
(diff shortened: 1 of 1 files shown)


src/components/InfoViewer/InfoViewer.scss
```diff
@@ -54,6 +54,8 @@
     &__value {
         display: flex;
 
+        min-width: 120px;
```
ariser, line 57: 120px, using the current font settings, makes Date standard string representation display in two rows, the first being date and the second being time. No other reason.

artemmufazalov (approved):

PR #259, opened 2022-12-21
chore: cover describe consumer and topic with types

artemmufazalov:


src/types/api/consumer.ts
```diff
@@ -0,0 +1,203 @@
...
+    consumer?: Consumer;
+    partitions?: PartitionInfo[];
+}
+
+// Partition info types differs for consumer and topic, although they are very similar
+export interface PartitionInfo {
+    /** int64  */
+    partition_id?: string;
+
+    // Is partition open for write.
```
ariser, line 18: I suppose these annotations can as well be jsdoc, to appear in autocomplete popups

ariser (approved):

src/types/api/consumer.ts
artemmufazalov: I agree, this is a very good idea. I will do that

ariser (approved):

PR #261, opened 2022-12-23
refactor(TopQueries): migrate to ts

ariser:
No difference in functionality, refactoring js-to-ts only


src/containers/Tenant/Diagnostics/TopQueries/TopQueries.tsx
```diff
@@ -0,0 +1,125 @@
...
+        (row) => {
+            const {QueryText: input} = row;
+
+            dispatch(changeUserInput({input}));
+            changeSchemaTab(TenantGeneralTabsIds.query);
+        },
+        [changeSchemaTab, dispatch],
+    );
+
+    const renderLoader = () => {
```
Raubzeug, line 86: I'm not sure 100%, but it seems we already have created specific Loader component for our app.

Raubzeug (approved):

src/containers/Tenant/Diagnostics/TopQueries/TopQueries.tsx
ariser: You're right, thanks. I will probably address this in the following PR.

PR #263, opened 2022-12-29
refactor(Tablets): migrate to ts

artemmufazalov:


src/containers/Tablets/Tablets.tsx
```diff
@@ -0,0 +1,166 @@
...
+        );
+    };
+
+    const renderContent = () => {
+        const states = Array.from(new Set(...[tablets.map((tablet) => tablet.State)]))
+            .filter((state): state is ETabletState => state !== undefined)
+            .map((item) => ({
+                value: item,
+                content: item,
+            }));
```
ariser, line 118: it was 5 traversals: map, spread, set, array.from, map
now it is 6: everything above + filter

I agree that migration from js to ts shouldn't alter the existing code too much, but let's at least remove the spread, it makes no sense

ariser (changes requested):

ariser (approved):

PR #265, opened 2022-12-29
refactor(Heatmap): migrate to ts

artemmufazalov:


src/store/reducers/heatmap.ts
```diff
@@ -47,16 +57,16 @@ const tablets = function z(state = initialState, action) {
...
             window.api.getTabletsInfo({nodes, path}),
             window.api.getHeatmapData({path}),
         ]),
-        actions: FETCH_TABLETS,
-        dataHandler: ([tabletsData = [], describe = {}]) => {
+        actions: FETCH_HEATMAP,
+        dataHandler: ([tabletsData = {}, describe = {}]) => {
             const {TabletStateInfo: tablets = []} = tabletsData;
-            const TabletsMap = new Map();
+            const TabletsMap: Map<string | undefined, IHeatmapTabletData> = new Map();
```
ariser, line 69: I'd say it's better to explicitly skip empty ids, and keep the keys type consistent as a string

src/containers/Heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+    const renderContent = () => {
+        const {min, max} = getCurrentMetricLimits(currentMetric, tablets);
+
+        return (
+            <div className={b()}>
+                <div className={b('filters')}>
+                    <Select
+                        className={b('heatmap-select')}
+                        value={[currentMetric || '']}
```
ariser, line 150: this recreates the value entity every render, changing its reference
judging by the Select sources, this only affects memoization of event handling functions inside the lib
still, not a very reliable approach

ariser (changes requested):

src/containers/Heatmap/Heatmap.tsx
```diff
@@ -0,0 +1,203 @@
...
+        wasLoaded,
+        error,
+        sort,
+        heatmap,
+        metrics,
+        currentMetric,
+        data: tablets = [],
+    } = useTypedSelector((state) => state.heatmap);
+
+    const [selectedMetric, setSelectedMetric] = useState(['']);
```
ariser, line 45: i was thinking more of useMemo than a state. But it works as well, so

ariser (approved):

PR #267, opened 2022-12-29
refactor(Storage): make EntitiesCount a component

artemmufazalov:


src/components/EntitiesCount/EntitiesCount.tsx
```diff
@@ -0,0 +1,37 @@
+import {Label} from '@gravity-ui/uikit';
+
+interface EntitiesCountProps {
+    countTotal: number | string;
+    countCurrent?: number | string;
```
ariser, line 5: seems to me that total should be optional, not current
the current counter is always available thru the collection size, while the total can be unknown

src/components/EntitiesCount/EntitiesCount.tsx
```diff
@@ -0,0 +1,37 @@
...
+    entityName?: string;
+    isLoadingStatus?: boolean;
+    className?: string;
```
ariser, line 8: verbose naming
```
total
current
name/title/label
loading
```

loading instead of isLoading because it's a prop, not a variable
```
<EntitiesCount loading />
```
vs
```
<EntitiesCount isLoading />
```

src/components/EntitiesCount/EntitiesCount.tsx
ariser: even in the current use case code can be simplified if you swap the optionals

src/containers/Storage/Storage.js
```diff
@@ -230,22 +231,20 @@ class Storage extends React.Component {
...
-            label += `${flatListStorageEntities.length} of ${count.total}`;
-        }
+        const countTotal = count.total ? count.total : flatListStorageEntities.length;
 
-        return label;
+        return (
+            <EntitiesCount
+                entityName={entityName}
+                isLoadingStatus={loading && !wasLoaded}
+                countTotal={countTotal}
```
ariser, line 244: just make it optional and pass `count.total`

ariser (changes requested):

src/components/EntitiesCount/EntitiesCount.tsx
ariser: although i don't really like loading state in this component. This is a counter, it shouldn't know about loading state, it doesn't load anything
I understand why, but I suggest look for another way

src/components/EntitiesCount/EntitiesCount.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (entityName) {
+        label += `${entityName}: `;
+    }
+
+    if (isLoadingStatus) {
+        label += '...';
+    } else if (countCurrent === undefined || Number(countCurrent) === Number(countTotal)) {
+        label += `${countTotal}`;
+    } else {
+        label += `${countCurrent} of ${countTotal}`;
```
ariser, line 29: i18n?

ariser (approved):

src/components/EntitiesCount/EntitiesCount.tsx
```diff
@@ -0,0 +1,34 @@
+import {Label} from '@gravity-ui/uikit';
+import i18n from './i18n';
+
+interface EntitiesCountProps {
+    total: number | string;
+    current?: number | string;
```
ariser, line 6: didn't swap the question mark

ariser (changes requested):

ariser (approved):

PR #269, opened 2022-12-30
fix(UserSettings): treat invertedDisks settings as string

ariser:


src/store/reducers/settings.js
```diff
@@ -30,7 +30,7 @@ export const initialState = {
         ...defaultUserSettings,
         ...userSettings,
         [THEME_KEY]: readSavedSettingsValue(THEME_KEY, 'light'),
-        [INVERTED_DISKS_KEY]: readSavedSettingsValue(INVERTED_DISKS_KEY) === 'true',
+        [INVERTED_DISKS_KEY]: readSavedSettingsValue(INVERTED_DISKS_KEY, 'false'),
```
artemmufazalov, line 33: We have `defaultUserSettings` in constants with default theme and inverted disks values. Probably it will be better to use them as default values in `readSavedSettingsValue`

artemmufazalov (changes requested):

src/store/reducers/settings.js
artemmufazalov: I noticed that you deleted default settings in other PR. So, it seems resolved)

artemmufazalov (approved):

PR #270, opened 2022-12-30
feat(Nodes): display rack in table

ariser:


artemmufazalov (approved):

artemmufazalov (approved):

src/containers/Storage/StorageNodes/StorageNodes.tsx
```diff
@@ -96,6 +98,12 @@ function StorageNodes({
             },
             align: DataTable.LEFT,
         },
+        {
+            name: TableColumnsIds.Rack,
+            header: tableColumnsNames[TableColumnsIds.Rack],
+            render: ({value}) => value as string || '—',
```
artemmufazalov, line 104: Unfortunately, react-data-table cannot get type of value by property name and set value type as unknown. In my nodes PR I was changing such things to  `({row}) => {...}` when added columns type (`Column<T>`). So it is better to extract value from `row` to make adding types easier.

artemmufazalov (changes requested):

src/containers/Storage/StorageNodes/StorageNodes.tsx
ariser: Makes sense. Will change it back to `row` to match your approach.
I believe this inconvenience will build up to me making a PR to react-data-table :)

artemmufazalov (approved):

PR #280, opened 2023-01-19
feat(Overview): display topic stats for topics and streams

artemmufazalov:


src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -122,6 +120,10 @@ function Overview({type, tenantName, className}: OverviewProps) {
                 }
                 dispatch(getOlapStats({path: schemaPath}));
             }
+
+            if (isPathTypeWithTopic(type)) {
+                dispatch(getTopic(currentSchemaPath));
```
artemmufazalov, line 125: Topic data is requested here because of loading and isLoading flags. When these flags are reset, overview child components are mounted again, so topic data is loaded twice if it is fetched inside child components

src/components/InfoViewer/utils.ts
```diff
@@ -40,3 +40,13 @@ export function createInfoFormatter<Shape extends Record<string, any>>({
...
+
+export const formatObject = (formatter: ReturnType<typeof createInfoFormatter>, obj?: Object) => {
+    if (!obj) {
+        return [];
+    }
+
+    return Object.entries(obj)
+        .map(([label, value]) => formatter(label, value))
+        .filter(({value}) => Boolean(value));
+};
```
ariser, line 52: this can be strictly typed

src/containers/Tenant/Diagnostics/Overview/ChangefeedInfo/ChangefeedInfo.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return (
+        <div>
+            <div>
+                <InfoViewer title={entityName} info={prepareChangefeedInfo(data, pqGroupData)} />
+            </div>
+            <div>
+                <TopicStats />
+            </div>
+        </div>
```
ariser, line 42: why so many divs?

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.scss
```diff
@@ -0,0 +1,33 @@
+.ydb-overview-topic-stats {
+    .ydb-loader {
+        margin-top: 50px;
+    }
+
+    &__info {
+        .info-viewer {
```
ariser, line 7: if you can define block styles without referencing other blocks, better do this

works:
```
<div className="block1">
    <div className="block1__elem">
        <div className="block2" />
    </div>
</div>

.block1__elem .block2 {
}
```

works better:
```
<div className="block1">
    <div className="block1__elem block2" />
</div>

.block1__elem {
}
```

or even if you need nesting for some reason
```
<div className="block1">
    <div className="block1__elem1">
        <div className="block1__elem2 block2" />
    </div>
</div>

.block1__elem2 {
}
```

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (
+        error ||
+        !data ||
+        typeof data !== 'object' ||
+        !data.topic_stats ||
+        isEmpty(data.topic_stats)
+    ) {
+        return (
+            <div>
+                <div className="info-viewer info-viewer__title">Stats</div>
```
ariser, line 77: you use block classname outside the block
it has no connection to the block, which makes it fragile
this will break when you decide to change InfoViewer classname for whatever reason. Or it will break if InfoViewer is no longer used, and its css in no longer included in bundle. And there is no way of preventing or detecting that

if you want to use block classname, extend or alter the block functionality so that the block itself would render this
if you want to mimic block appearance, maybe mixins could help, we use scss. Or maybe you don't want to mimic, just repeat the styles here, them matching can be a coincidence

src/containers/Tenant/Diagnostics/Overview/TopicStats/i18n/ru.json
```diff
@@ -0,0 +1,3 @@
+{
+  "notSupportedVersion": "Статистика о топиках не поддерживается на текущей версии. Необходимо обновить кластер до более новой версии"
```
ariser, line 2: Статистика топиков?
в текущей версии?

Необходимо обновить → Обновите <...>, чтобы <...>

ariser: может быть мы можем написать, в какой версии есть нужная функциональность

src/containers/Tenant/utils/schema.ts
```diff
@@ -217,3 +217,22 @@ const pathTypeToChildless: Record<EPathType, boolean> = {
 
 export const isChildlessPathType = (type?: EPathType, subType?: EPathSubType) =>
     ((subType && pathSubTypeToChildless[subType]) || (type && pathTypeToChildless[type])) ?? false;
+
+// ====================
+
+const mapPathTypeToWithTopic: Record<EPathType, boolean> = {
```
ariser, line 223: ToIsWithTopic?

src/containers/Tenant/Diagnostics/Overview/utils/prepareChangefeedInfo.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import {formatObject, InfoViewerItem} from '../../../../../components/InfoViewer';
+import {
+    formatCdcStreamItem,
+    formatCommonItem,
+} from '../../../../../components/InfoViewer/formatters';
+
+import {prepareTopicSchemaInfo} from './prepareTopicSchemaInfo';
+
+export const prepareChangefeedInfo = (
```
ariser, line 11: this was in CDCStreamInfo.tsx, you moved it in utils, correct?
Do you expect this to be a shared function that would be required in other components? If yes, ok, if not — why do we need utils for this?

ariser (changes requested):

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.scss
ariser: an example to compare assumptions
```
.block {
    // I have an element inside me, it is implemented using block ydb-text, its color id red
    .ydb-text {
        color: red;
    }
}

.block {
    // I have an element A, its color is red
    &__text {
        color: red;
    }
}
```

src/containers/App/App.js
```diff
@@ -14,6 +14,10 @@ import {registerLanguages} from '../../utils/monaco';
 
 import './App.scss';
 
+i18n.setLang(Lang.En);
+configureYdbUiComponents({lang: Lang.En});
+configureUiKit({lang: Lang.En});
```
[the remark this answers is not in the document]
artemmufazalov: I don't understand your point. Do you mean moving this to configuration file to ensure correct setup for every import?

src/styles/mixins.scss
```diff
@@ -152,6 +149,17 @@
     .data-table__row:hover .data-table__td:nth-child(#{$nth}) {
         background-color: var(--ydb-data-table-color-hover) !important;
     }
+
+    @for $i from 0 through $nth {
```
artemmufazalov, line 153: Stylelint refused to commit without this fix (expected rule to come before at-rule with a block)

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.scss
```diff
@@ -0,0 +1,38 @@
...
+
+    &__info {
+        .info-viewer__label-text {
+            display: flex;
+            overflow: visible;
+
+            max-width: 150px;
+
+            white-space: normal;
+        }
```
ariser, line 24: in my previous story about defining styles for an element and not for a nested block I didn't realize this was styles for an element of info-viewer, not for the root of info-viewer.

but now I have another question: are these styles to make labels multiline?
seems like it should be an InfoViewer internal behaviour enabled by a prop, not an outside modification

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.scss
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    &__bytes-written {
+        margin-top: 7px;
+
+        .info-viewer {
+            padding-left: 20px;
```
ariser, line 31: this is however _is_ a reference to another block, which could be a style for an element of the current block

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.scss
```diff
@@ -0,0 +1,38 @@
...
+.ydb-overview-topic-stats {
+    &__title {
+        @include info-viewer-title();
+    }
+
+    .ydb-loader {
+        margin-top: 50px;
+    }
+
+    .ydb-bars {
```
ariser, line 12: as is this

ariser (changes requested):

ariser (approved):

PR #286, opened 2023-01-20
feat(Storage): display vdisks over pdisks

ariser:
based on #285
if reviewing before its merge, look only at the last commit

### DIFF
--- a/src/containers/Storage/PDisk/PDisk.tsx
+++ b/src/containers/Storage/PDisk/PDisk.tsx
@@ -4,13 +4,18 @@ import cn from 'bem-cn-lite';
 import {InternalLink} from '../../../components/InternalLink';
 
 import routes, {createHref} from '../../../routes';
+import {getVDisksForPDisk} from '../../../store/reducers/storage';
 import {TPDiskStateInfo, TPDiskState} from '../../../types/api/pdisk';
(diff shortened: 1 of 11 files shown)


src/containers/Storage/PDisk/PDisk.tsx
```diff
@@ -4,17 +4,22 @@ import cn from 'bem-cn-lite';
...
 import {STRUCTURE} from '../../Node/NodePages';
 
 import {DiskStateProgressBar, EDiskStateSeverity} from '../DiskStateProgressBar';
 import {PDiskPopup} from '../PDiskPopup';
+import {VDisk} from '../VDisk';
 
 import {NOT_AVAILABLE_SEVERITY} from '../utils';
 
 import './PDisk.scss';
+import {TVDiskStateInfo} from '../../../types/api/vdisk';
```
artemmufazalov, line 22: It will be better to move this import higher)

artemmufazalov (changes requested):

artemmufazalov:
In case, when a PDisk in a row doesn't have VDisks, while others do, it align to the top. In my view, it's better for the to be aligned to the bottom
Screen Shot 2023-01-23 at 18 37 03

artemmufazalov (approved):

PR #288, opened 2023-01-24
refactor(SchemaInfoViewer): migrate to ts and rename

artemmufazalov:


src/containers/Tenant/Diagnostics/Overview/Overview.scss
```diff
@@ -1,13 +1,35 @@
-.kv-tenant-overview {
+@import '../../../../styles/mixins.scss';
+
+.ydb-diagnostics-overview {
```
artemmufazalov, line 3: I want to reuse these styles for TableIndexInfo further, so I moved them on the level of Overview component

src/containers/Tenant/Diagnostics/Overview/utils/prepareTableInfo.ts
```diff
@@ -0,0 +1,96 @@
+import type {TEvDescribeSchemeResult} from '../../../../../types/api/schema';
```
artemmufazalov, line 1: PrepareTableInfo will be reused for TableIndexComponent

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <div className={b()}>
+            <InfoViewer info={generalTableInfo} title={entityName} className={b('info-block')} />
+            <div className={b('row')}>
+                {tabletMetricsInfo.length > 0 || partitionConfigInfo.length > 0 ? (
+                    <div className={b('col')}>
+                        <InfoViewer
+                            info={tabletMetricsInfo}
+                            title={'Tablet Metrics'}
+                            className={b('info-block')}
+                            nullOnEmptyInfo
```
artemmufazalov, line 47: Previously TableInfo (ex. SchemaInfoViewer) had function helper `renderItem`, that returned null on empty info. I decided to replace it with a new prop

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import '../Overview.scss';
+
+const b = cn('ydb-diagnostics-overview');
+
+interface TableInfoProps {
+    data?: TEvDescribeSchemeResult;
+}
+
+export const TableInfo = ({data}: TableInfoProps) => {
```
artemmufazalov, line 20: I renamed SchemaInfoViewer to TableInfo as it designed to process table data mostly

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -170,15 +158,15 @@ function Overview({type, tenantName, className}: OverviewProps) {
...
-        );
+        return (type && pathTypeToComponent[type]?.()) || <TableInfo data={schemaData} />;
     };
 
     if ((loading && !wasLoaded) || (isEntityWithMergedImpl && !mergedChildrenPaths)) {
-        return renderLoader();
+        return <Loader size="m" />;
+    }
+
+    if (error) {
```
artemmufazalov, line 168: Previously, error wasn't displayed in Overview except Topics and Changefeeds. Instead there were "No entityName data" inside SchemaInfoViewer. It seems, in some cases it's better to see actual error

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -170,15 +158,15 @@ function Overview({type, tenantName, className}: OverviewProps) {
...
+        return <div className="error">{error.statusText}</div>;
```
artemmufazalov, line 169: Previously, error wasn't displayed in Overview except Topics and Changefeeds. Instead there were "No entityName data" inside SchemaInfoViewer. It seems, in some cases it's better to see actual error

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -15,6 +15,7 @@ interface InfoViewerProps {
     size?: 's';
     className?: string;
     multilineLabels?: boolean;
+    nullOnEmptyInfo?: boolean;
```
Raubzeug, line 18: It seems it will be more useful if you add `renderEmptyState` prop - a function that returns a component. If prop is not passed, you return null.

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -81,6 +76,7 @@ function Overview({type, tenantName, className}: OverviewProps) {
         wasLoaded,
         autorefresh,
         currentSchemaPath,
+        error,
```
Raubzeug, line 79: Is there any error parser in project? If no, I think it's useful to add it.

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
Raubzeug, line 169: It would be better to use parsed error (I mentioned error parser in prev comment). Because as I remember it's a lot of error constructions in project.

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const entityName = getEntityName(data?.PathDescription);
+
+    const {
+        generalTableInfo = [],
+        tableStatsInfo = [],
+        tabletMetricsInfo = [],
+        partitionConfigInfo = [],
+    } = useMemo(() => prepareTableInfo(data), [data]);
+
+    if (
```
Raubzeug, line 30: let's put it into a helper function `checkIsEmptyTableInfo`.
Why `tableStatsInfo` is flatten, but other arrays are not?

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+                            info={partitionConfigInfo}
+                            title={'Partition Config'}
+                            className={b('info-block')}
+                            nullOnEmptyInfo
+                        />
+                    </div>
+                ) : null}
+                <div className={b('col')}>
+                    {tableStatsInfo.map((info, index) => (
+                        <React.Fragment key={index}>
```
Raubzeug, line 59: `React.Fragment` seems to be redundant

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    return (
+        <div className={b()}>
+            <InfoViewer info={generalTableInfo} title={entityName} className={b('info-block')} />
+            <div className={b('row')}>
+                {tabletMetricsInfo.length > 0 || partitionConfigInfo.length > 0 ? (
+                    <div className={b('col')}>
+                        <InfoViewer
+                            info={tabletMetricsInfo}
+                            title={'Tablet Metrics'}
```
Raubzeug, line 45: Shouldn't we use i18n throughout refactored components?

src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.ts
```diff
@@ -0,0 +1,96 @@
+import type {TEvDescribeSchemeResult} from '../../../../../types/api/schema';
+
+import {formatObject} from '../../../../../components/InfoViewer';
```
Raubzeug, line 3: if the function is used more than in one component, let's move it to `helpers`.

Raubzeug (changes requested):

src/components/InfoViewer/InfoViewer.tsx
artemmufazalov: By default component returns title and no data error message. But the idea with the functionÏ looks very good, this prop may be very useful in future

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
artemmufazalov: There is a function `prepareQueryError`, that was used for `sendQuery` requests. I used it before but it doesn’t fit well as for me

Now I added `ResponseError` component, which wraps error with `IResponseError` type

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
artemmufazalov: `tableStatsInfo` is flatten, because it is array of arrays (`[][]`).

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
artemmufazalov: Explicit `React.Fragment` makes possible to add key prop to it

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
artemmufazalov: I doubted whether to use i18n for titles or not because in some places titles contain entity names that currently have mapping only in English. So it will be the mix of Russian and English. Now I added translations, It’s better to have them at least somewhere

src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.ts
artemmufazalov: Currently it is not used somewhere apart from TableInfo. That’s why I placed it here

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
Raubzeug: Maybe it make sense to join all error parsers in one, not to think what type of error do you deal with?

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export const TableInfo = ({data}: TableInfoProps) => {
+    const entityName = getEntityName(data?.PathDescription);
+
+    const {
+        generalTableInfo = [],
+        tableStatsInfo = [],
+        tabletMetricsInfo = [],
+        partitionConfigInfo = [],
+    } = useMemo(() => prepareTableInfo(data), [data]);
```
Raubzeug, line 42: Is `useMemo` really needed here?

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
Raubzeug: It is valid to add `key` prop directly to `InfoViewer`

src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.ts
Raubzeug: I meant that you imported {formatObject} from '../../../../../components/InfoViewer' . If this function is used only in TableInfo, maybe you should move it to `prepareTableInfo.ts`?

Raubzeug (changes requested):

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -26,28 +27,35 @@ const InfoViewer = ({
...
-    <div className={b({size}, className)}>
-        {title && <div className={b('title')}>{title}</div>}
-        {info && info.length > 0 ? (
-            <div className={b('items')}>
-                {info.map((data, infoIndex) => (
-                    <div className={b('row')} key={data.label + infoIndex}>
-                        <div className={b('label')}>
-                            <div className={b('label-text', {multiline: multilineLabels})}>
-                                {data.label}
+    renderOnEmptyState,
```
Raubzeug, line 30: `renderOnEmptyState = () => null` not to provide this prop if you just want to render null (it seems it's a common case).

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -15,6 +15,7 @@ interface InfoViewerProps {
     size?: 's';
     className?: string;
     multilineLabels?: boolean;
+    renderOnEmptyState?: () => JSX.Element | null;
```
ariser, line 18: i believe the suggested prop name was `renderEmptyState` (not `...On...`)

src/components/InfoViewer/InfoViewer.tsx
ariser: no, the common case is to render `No ${title} data`

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -26,28 +27,35 @@ const InfoViewer = ({
...
-            <div className={b('items')}>
-                {info.map((data, infoIndex) => (
-                    <div className={b('row')} key={data.label + infoIndex}>
-                        <div className={b('label')}>
-                            <div className={b('label-text', {multiline: multilineLabels})}>
-                                {data.label}
+    renderOnEmptyState,
+}: InfoViewerProps) => {
+    if ((!info || !info.length) && renderOnEmptyState) {
+        return renderOnEmptyState();
```
ariser, line 33: i'd pass displayable data there
surely title
maybe size

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import {prepareTableInfo} from './prepareTableInfo';
+
+import i18n from './i18n';
+
+import './TableInfo.scss';
+
+const b = cn('ydb-diagnostics-table-info');
+
+const checkIsEmptyTableInfo = (
```
ariser, line 18: why `check...`?
just `is...`?

ariser: also, well

“is table info empty”

not 

“is empty table info”

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
ariser: i'd say yes. `prepareTableInfo` is a complex function

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
```diff
@@ -0,0 +1,90 @@
...
+        partitionConfigInfo = [],
+    } = useMemo(() => prepareTableInfo(data), [data]);
+
+    if (
+        checkIsEmptyTableInfo(
+            generalTableInfo,
+            tabletMetricsInfo,
+            partitionConfigInfo,
+            tableStatsInfo,
+        )
```
ariser, line 50: don't have any particular solution in mind, just notice that this still looks heavy
also — what if the set of fields determining the emptiness changes? This will entail changes here as well as in the function itself
i'd like to write something like `if (isEmpty(data)) {}` here and never touch it again
wouldn't have the fields calculated twice though, of course

src/components/Errors/ResponseError/ResponseError.tsx
```diff
@@ -0,0 +1,11 @@
+import type {IResponseError} from '../../../types/api/error';
+
+interface ResponseErrorProps {
+    error: IResponseError;
+    className?: string;
+    defaultMessage?: string;
```
ariser, line 6: maybe default to something like “Response error”? just to have anything for an empty input

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
ariser: a know-it-all parser will become a mess sooner or later
ideally a component should know what errors it have to deal with
and there can be a common parser/renderer to default to

ariser (changes requested):

src/components/InfoViewer/InfoViewer.tsx
ariser: it should also just return `ReactNode`, which includes null, and will allow a simple string

src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.ts
artemmufazalov: Somehow I supposed your question was about `prepareTableInfo`. So my previous comment was not about `formatObject`.

`formatObject` is also used in 3 other components: `TopicInfo`, `ChangefeedInfo` and `TopicStats`. This function applies formatter, which is return type of `createInfoFormatter`, to object. So I placed it where `createInfoFormatter`, `formatLabel`, `formatValue` functions and all connected types were declared. All formatters, which can be passed to `formatObject`, are also declared in components/InfoViewer/formatters.

src/components/InfoViewer/InfoViewer.tsx
artemmufazalov: What do you mean by displayable data here? If you mean `info`, what the sense in passing it here if it is an empty array or undefined?

src/containers/Tenant/Diagnostics/Overview/TableInfo/TableInfo.tsx
artemmufazalov: I discovered, that we don’t need such a check here. We can just add renderEmptyState on `generalTableInfo` block. In case if every field is empty, it will render a title, as every other `renderEmptyState` return null. Also it tackles the case, when `generalTableInfo` block is empty, while others aren’t (it will be displayed “No Table data” while there will be some data in other blocks)

src/components/InfoViewer/InfoViewer.tsx
ariser: surely title
maybe size

ariser (approved):

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
Raubzeug: Thanks, agree with you!

src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.ts
Raubzeug: Maybe it will be more comfortable to move all this formatters to `utils/formatters`. It's just pondering, not request for immediate changes.

Raubzeug (approved):

PR #294, opened 2023-01-31
fix: minor fixes in Nodes and Tenants tables

artemmufazalov:


src/containers/Tenants/Tenants.js
```diff
@@ -218,8 +219,8 @@ class Tenants extends React.Component {
                 },
                 accessor: ({Metrics = {}, CoresUsed}) => {
                     if (!isNaN(Number(CoresUsed))) {
-                        const cores = Math.round(Number(CoresUsed) * 100) / 100;
-                        return cores || '—';
+                        const cores = Number(CoresUsed).toFixed(2);
+                        return Number(cores) ? numeral(cores).format('0.00') : '—';
```
ariser, line 223: what is the purpose of this change?
maybe extend [formatNumber](https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/utils/index.js#L58) with a precision argument and use it here?
there is also [formatCPU](https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/utils/index.js#L66), but looks like the orders of values don't match

artemmufazalov: The purpose of this change to keep 2 values after a dot and display dash for zero values. I'm not sure that it makes sense to move that logic (display dash if zero) somewhere else

artemmufazalov: The difference is that now it always displays 2 values after a dot, previously it was just rounded

ariser: dash should remain here, of course
i'm talking about the formatting, there is a function that does that. It just lacks the precision argument, to specify these 2 fraction values. And it handles non-numeric values better than `if (Number(value)) {}`, you can just `formatNumber(value) || '—'`

ariser: ah, zero should be a dash, i see

ariser: my concern is currently `numeral` is only used in utils, and the rest of the app uses functions from utils to format values, this makes dependency change easy
in this PR you bring `numeral` to another module

ariser: so all the logic that is going on here is fine
my comment is about formatting specifically

ariser (approved):

PR #300, opened 2023-02-02
fix(Auth): add a step in history for auth form

ariser:


artemmufazalov:
Why did you decide to keep `setIsNotAuthenticated` and `setIsAuthenticated` functions, although they are not used anymore?

ariser:
> Why did you decide to keep setIsNotAuthenticated and setIsAuthenticated functions, although they are not used anymore?

To keep backwards compatibility. They are still imported in the internal version. Will remove them after we update the internal code.

artemmufazalov (approved):

PR #301, opened 2023-02-02
fix(QueryEditor): better report timeout errors

ariser:


src/containers/Tenant/QueryEditor/QueryResult/QueryResult.js
```diff
@@ -94,31 +94,33 @@ function QueryResult(props) {
...
-                            <ResultIssues data={error} />
+                            <ResultIssues data={error.data} />
                         </div>
                     </Fullscreen>
                 )}
             </React.Fragment>
         ) : (
-            <span>{error?.data ?? error}</span>
+            <div className={b('error')}>
+                {error?.message || error?.data || error}
```
artemmufazalov, line 114: Can error data be an object? Or error itself? Won't it fail with "Objects are not valid as a React child"? Maybe it's better to use `prepareQueryError` as it seems to fit here. Or at least parse it to string

src/services/api.js
```diff
@@ -185,7 +185,10 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
                 timeout: 600000,
```
artemmufazalov, line 185: Is it a backend timeout? Why 2 timeouts?

src/components/QueryExecutionStatus/QueryExecutionStatus.tsx
```diff
@@ -1,29 +1,46 @@
...
+import questionIcon from '../../assets/icons/question.svg';
+
 import './QueryExecutionStatus.scss';
 
 const b = cn('kv-query-execution-status');
 
 interface QueryExecutionStatusProps {
     className?: string;
-    hasError?: boolean;
+    error?: Record<string, any>;
```
artemmufazalov, line 16: I think It is better to introduce some interface `IQueryResponseError` and at least specify  properties that could be present. It makes it much easier to understand

src/services/api.js
ariser: one for backend operations, one for the client request

src/containers/Tenant/QueryEditor/QueryResult/QueryResult.js
```diff
@@ -94,31 +100,33 @@ function QueryResult(props) {
...
-                            <ResultIssues data={error} />
+                            <ResultIssues data={error.data} />
                         </div>
                     </Fullscreen>
                 )}
             </React.Fragment>
         ) : (
-            <span>{error?.data ?? error}</span>
+            <div className={b('error')}>
+                {prepareQueryError(error)}
```
artemmufazalov, line 120: It seems error could be undefined here. And `prepareQueryError` expects error to be defined. 

https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/utils/query.ts#L188

ariser: fixed

artemmufazalov (dismissed):

artemmufazalov (approved):

PR #302, opened 2023-02-02
fix: format dates with date-utils

ariser:


artemmufazalov:
I don't really understand the reason but new time is broken for Changefeeds and Topics.

### Before:
Screen Shot 2023-02-03 at 15 59 54

### Now (local docker):
Screen Shot 2023-02-03 at 16 00 57

### Now (real cluster):
Screen Shot 2023-02-03 at 16 04 50

ariser:
fixed

artemmufazalov (approved):

PR #304, opened 2023-02-03
chore: move to @gravity-ui/react-data-table

artemmufazalov:


ariser:
Something wrong in ci

config-overrides.js
```diff
@@ -20,8 +23,12 @@ module.exports = {
         return config;
     },
     jest: (config) => {
+        // Some @gravity-ui libs (react-data-table, paranoid) are build in esm only
+        // So they need to be transformed
+        // By default jest does not transform anything in node_modules
+        // So this override excludes node_modules except @gravity-ui
```
ariser, line 29: it's the other way around
this option excludes the path from being transformed
cuz if it's not, the transformation fails, since they are already esm, and babel expects cjs

ariser: > this option excludes the path from being transformed

not the option itself, but this particular regexp

artemmufazalov: The pattern is `node_modules/(?!(@gravity-ui)/)`, so `@gravity-ui` is actually excluded from `transformIgnorePatterns`. As I understand, `transformIgnorePatterns` determine paths, that are not transformed. So files in path`/node_modules/@gravity-ui/` are transformed

ariser: no, my bad, the current comment is correct

ariser (approved):

PR #307, opened 2023-02-07
fix(TopShards): switch between history and immediate data

ariser:


src/utils/typecheckers.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Intended to typecheck enums, but also checks if the value is any of the arbitrary object values
+ */
+export const isEnum = <T extends Object>(object: T, value: any): value is T[keyof T] =>
```
artemmufazalov, line 4: I don't like the naming here, for me it is not clear, what this function does. I suppose something like `isEnumValue` or `isValueInEnum` will be more explicit

src/containers/Tenant/Diagnostics/TopShards/TopShards.tsx
```diff
@@ -231,7 +257,7 @@ export const OverloadedShards = ({tenantPath, type}: OverloadedShardsProps) => {
             {
                 name: tableColumnsNames.IntervalEnd,
                 render: ({value}) => formatDateTime(new Date(value as string).getTime()),
```
artemmufazalov, line 259: If it is `immediate` table, this fields (`IntervalEnd` and `PeakTime`) are missing. I think here it's needed to add some placeholders like dashes or somehow else say, that these fields are always empty for this mode

src/containers/Tenant/Diagnostics/TopShards/Filters/Filters.tsx
```diff
@@ -0,0 +1,56 @@
...
+import './Filters.scss';
+
+interface FiltersProps {
+    value: IShardsWorkloadFilters;
+    onChange: (value: Partial<IShardsWorkloadFilters>) => void;
+    className?: string;
+}
+
+export const Filters = ({value, onChange, className}: FiltersProps) => {
+    const handleModeChange = (mode: string) => {
```
artemmufazalov, line 24: What is the reasons of a such check, if you define both filters and handler in the same component?

src/containers/Tenant/Diagnostics/TopShards/Filters/Filters.tsx
```diff
@@ -0,0 +1,56 @@
...
+            mode: EShardsWorkloadMode.History,
+            ...dateRange,
+        });
+    };
+
+    const from = value.mode === EShardsWorkloadMode.Immediate ? undefined : value.from;
+    const to = value.mode === EShardsWorkloadMode.Immediate ? undefined : value.to;
+
+    return (
+        <div className={b('filters', className)}>
```
artemmufazalov, line 44: Will it be better not to import `b` here, but pass className `b('filters')` via props as your styles are defined on TopShards level?

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/TopShards/Filters/Filters.tsx
ariser: `RadioButton` doesn't allow specifying its value type, it's just `string`
`onChange` expects a restricted filters shape with `mode` being of type `EShardsWorkloadMode`
`string` can't be assigned to `EShardsWorkloadMode`
this check narrows incoming value type

src/containers/Tenant/Diagnostics/TopShards/Filters/Filters.tsx
ariser: no, it's this element name, and its styles are in Filters.scss

src/containers/Tenant/Diagnostics/TopShards/TopShards.tsx
ariser: i thought i addressed it
thanks
will probably remove these columns entirely for the immediate mode

ariser: fixed

artemmufazalov (approved):

PR #320, opened 2023-02-20
refactor(Icon): migrate to ts

artemmufazalov:


src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,32 @@
...
+    width = 16,
+    viewBox = '0 0 16 16',
+    className,
+    onClick,
+}: IconProps) => {
+    return (
+        <UiKitIcon
+            // @ts-expect-error
+            // Both url and id strings are required here.
+            // However, if no url provided component uses id, what is needed
```
ariser, line 24: had trouble understanding this
what would help:
- `here` → `in the uikit component`
- `what is needed` → `which is what we need` or smth

also, `@ts-expect-error` inverts exceptions behaviour, and will throw once this issue is fixed in uikit (I assume this is a bug in the component typings). A tip for the future developer who will encounter this exception would be helpful, e.g. "if fixed in uikit, can safely remove"

ariser (changes requested):

ariser (dismissed):

ariser (approved):

PR #321, opened 2023-02-21
refactor: migrate some components to ts

artemmufazalov:


src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,22 @@
+import cn from 'bem-cn-lite';
+
+import {Tag, TagType} from '../Tag/Tag';
```
ariser, line 3: from '../Tag'?

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+interface TagsProps {
+    tags: string[];
+    tagsType?: TagType;
+    className?: string;
+}
+
+export const Tags = ({tags, tagsType, className = ''}: TagsProps) => {
+    return (
+        <div className={`${b()} ${className}`}>
```
ariser, line 17: no need for the `className` default value, and the mix is applied as `b(null, className)` without strings concatenation

src/components/EmptyState/EmptyState.tsx
```diff
@@ -12,7 +13,15 @@ const sizes = {
     l: 350,
 };
 
-export default function EmptyState({image, title, description, actions, size}) {
+interface EmptyStateProps {
+    title: string;
+    image?: ReactNode;
+    description?: string | ReactNode;
```
ariser, line 19: ReactNode includes string

src/containers/Tablets/Tablets.tsx
```diff
@@ -104,7 +104,6 @@ export const Tablets = ({path, nodeId, className}: TabletsProps) => {
                 onMouseEnter={onShowTooltip}
                 tablet={tabletsToRender[tabletIndex]}
                 key={tabletIndex}
-                className={b('tablet')}
```
ariser, line 107: i believe there are styles for this classname in Tablets.scss?

src/components/Tablet/Tablet.tsx
```diff
@@ -0,0 +1,60 @@
+import {useRef} from 'react';
+import cn from 'bem-cn-lite';
+
+import type {TTabletStateInfo} from '../../types/api/tablet';
+
+import {InternalLink} from '../InternalLink';
+
+import {showTooltip} from '../../store/reducers/tooltip';
```
ariser, line 8: oof, don't really like importing a module just for its type
creates an excess edge in dependency graph
but i think this comment is just to bring attention to it, not to fix right away

ariser (changes requested):

src/containers/Tablets/Tablets.tsx
artemmufazalov: These styles wasn't applied in `Tablet`. So I removed prop but forgot to remove styles themselves.

ariser (approved):

PR #324, opened 2023-03-01
fix(Consumers): enable navigation to Partitions tab

artemmufazalov:


src/containers/Tenant/Diagnostics/Consumers/columns/columns.tsx
```diff
@@ -23,7 +29,27 @@ export const columns: Column<IPreparedConsumerData>[] = [
...
+            if (!row.name) {
+                return '–';
+            }
+
+            const queryParams = qs.parse(location.search, {
+                ignoreQueryPrefix: true,
+            });
+
+            return (
+                <EntityStatus
```
ariser, line 42: why EntityStatus? Isn't it just a link?

artemmufazalov: This component is used for links almost in every table (Tenants, Nodes, Partitions), so I used it to keep similar styles across the project

ariser: yea, but in all of those places there is a status icon and a clipboard button
this column only contains link

ariser: i mean, this is not an entity status, this is a link to a partition, no statuses here
in some of the cases where EntityStatus is being used now, it is also semantically incorrect. The solution is to either re-purpose the component and rename it, or create another one, or extract some functionality from it and create two components that would match all use cases semantically.
The issue I'm pointing at is semantics.

ariser: but right now there is a component `InternalLink` for example

artemmufazalov: Changed to `Link`

src/containers/Tenant/Diagnostics/Consumers/columns/columns.tsx
```diff
@@ -23,7 +29,28 @@ export const columns: Column<IPreparedConsumerData>[] = [
...
+                return '–';
+            }
+
+            const queryParams = qs.parse(location.search, {
+                ignoreQueryPrefix: true,
+            });
+
+            return (
+                <Link
+                    className="link"
```
ariser, line 43: what is this classname?
uikit classname is `yc-link`
`InternalLink` uses it

artemmufazalov: It is the classname defined in mixins

artemmufazalov: Changed to `InternalLink`

ariser: i don't think it is used anywhere directly, save the `@extend` statements

my issues with this:
- classname is determined without bem-cn-lite, which is used everywhere else (the `'error'` classname has the same issue btw)
- styles for this classname do not depend on uikit link styles, and can diverge

basically, my suggestion is to use `InternalLink`
it encapsulates uikit link styles combined with react-router-dom link behavior

ariser: that took me too long to phrase

PR #327, opened 2023-03-02
fix(PDisk): display vdisks donors

ariser:


src/containers/Storage/PDisk/PDisk.tsx
```diff
@@ -109,19 +111,42 @@ export const PDisk = ({nodeId, data: rawData = {}}: PDiskProps) => {
...
+                            {donors && donors.length ? (
+                                <Stack className={b('donors-stack')} key={stringifyVdiskId(vdisk.VDiskId)}>
+                                    <VDisk data={vdisk} compact />
+                                    {donors.map((donor) => {
+                                        const isFullData = isFullVDiksData(donor);
+
+                                        return (
+                                            <VDisk
+                                                compact
+                                                data={donor}
```
artemmufazalov, line 136: It seems that in this case not full donors will be without Donor label. It should be added manually via `DonorMode: true` property as it is in `StorageGroups`

src/containers/Storage/PDisk/PDisk.tsx
```diff
@@ -109,19 +111,42 @@ export const PDisk = ({nodeId, data: rawData = {}}: PDiskProps) => {
...
+                        <div
+                            key={stringifyVdiskId(vdisk.VDiskId)}
+                            className={b('vdisks-item')}
+                            style={{
+                                // 1 is small enough for empty disks to be of the minimum width
+                                // but if all of them are empty, `flex-grow: 1` would size them evenly
+                                flexGrow: Number(vdisk.AllocatedSize) || 1,
+                            }}
+                        >
+                            {donors && donors.length ? (
```
artemmufazalov, line 127: Not a request, but a suggestion. Since similar logic for processing and displaying VDisks and their donors is used in two places, maybe it makes sense to create some VDisksWrapper, that will incapsulate the logic?

artemmufazalov (changes requested):

src/containers/Storage/PDisk/PDisk.tsx
ariser: Not sure yet. They are similar, but still different enough to make a common component excessively complex

artemmufazalov (approved):

PR #331, opened 2023-03-06
fix: add opacity to unavailable nodes

artemmufazalov:


src/containers/Nodes/Nodes.tsx
```diff
@@ -166,6 +166,9 @@ export const Nodes = ({tenantPath, className, additionalNodesInfo = {}}: NodesPr
                             order: DataTable.ASCENDING,
                         }}
                         emptyDataMessage={i18n('empty.default')}
+                        rowClassName={(row) =>
+                            isUnavailableNode(row) ? b('unavailable-node') : ''
```
ariser, line 170: i'd expect it to be a modifier
```ts
b('node', {unavailable: isUnavailableNode(row)})
```
```scss
&__node_unavailable {
    opacity: 0.6;
}
```

src/types/api/nodes.ts
```diff
@@ -7,7 +7,7 @@ import {TVDiskStateInfo} from './vdisk';
 // source: https://github.com/ydb-platform/ydb/blob/main/ydb/core/viewer/protos/viewer.proto
 
 export interface TNodesInfo {
-    Overall: EFlag;
+    Overall?: EFlag;
```
ariser, line 10: is this a part of making unavailable nodes transparent?

src/containers/Nodes/Nodes.tsx
artemmufazalov: Thanks, it looks better. I'll change it

ariser (approved):

PR #337, opened 2023-03-22
fix(Storage): get nodes hosts from /nodelist

artemmufazalov:


src/store/reducers/nodesList.ts
```diff
@@ -36,13 +44,17 @@ const nodesList = function (state = initialState, action) {
...
     });
 }
 
+export const selectNodesMap = (state: NodesListRootStateSlice) =>
+    state.nodesList.data?.reduce<NodesMap>((nodesMap, node) => {
+        if (node.Id && node.Host) {
+            nodesMap[node.Id] = node.Host;
+        }
+        return nodesMap;
+    }, {});
```
Raubzeug, line 58: Let's use Map here, it is more efficient, when it is needed to change values inside.

Raubzeug (changes requested):

Raubzeug (approved):

PR #339, opened 2023-03-23
fix(TopicStats): use prepared stats, update fields

artemmufazalov:
Refactor topicStats. 
Change topicStats fields names, so they correspond with the names in Partitions and Consumers tables.
Add question icon with popover, that has explanation text and image.


src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.tsx
```diff
@@ -56,24 +81,14 @@ export const TopicStats = () => {
...
-    // 2. No data
-    // 3. HTML page of Internal Viewer with an error
-    // 4. Data with no topic stats
-    // 5. Topic Stats as an empty object
-    if (
-        error ||
-        !data ||
-        typeof data !== 'object' ||
-        !data.topic_stats ||
-        isEmpty(data.topic_stats)
```
artemmufazalov, line 71: Check for no data or data represented by raw text now is in reducer

src/containers/Tenant/Diagnostics/Overview/TopicStats/i18n/en.json
```diff
@@ -1,3 +1,4 @@
 {
-  "notSupportedVersion": "Topic stats are not supported in the current cluster version. Update cluster version to see topic stats"
```
artemmufazalov, line 2: Topics statistics are supported in ydb current stable version

src/containers/Tenant/Diagnostics/Overview/TopicStats/TopicStats.tsx
```diff
@@ -56,24 +81,14 @@ export const TopicStats = () => {
...
-    ) {
+    // If there is at least some empty data object
+    // we initialize its fields with zero values
+    // so no data at all is considered to be error as well
+    if (error || !data) {
         return (
             <div className={b()}>
                 <div className={b('title')}>Stats</div>
-                <div className="error">{i18n('notSupportedVersion')}</div>
```
artemmufazalov, line 76: Topics statistics are supported in ydb current stable version

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -39,7 +39,7 @@ const InfoViewer = ({
             {info && info.length > 0 ? (
                 <div className={b('items')}>
                     {info.map((data, infoIndex) => (
-                        <div className={b('row')} key={data.label + infoIndex}>
+                        <div className={b('row')} key={infoIndex}>
```
Raubzeug, line 42: It's a bad practice to generate keys on the fly: https://react.dev/learn/rendering-lists#rules-of-keys

src/components/LagPopover/LagPopover.scss
```diff
@@ -0,0 +1,13 @@
+.ydb-lag-popover {
+    div:nth-child(1) {
```
Raubzeug, line 2: Let's define css properties only with classes, not with html-tags, it's not a BEM-way.

src/components/LagPopover/LagPopover.tsx
```diff
@@ -0,0 +1,19 @@
...
+import './LagPopover.scss';
+
+const b = block('ydb-lag-popover');
+
+interface LagPopoverProps {
+    text: string;
+    type: 'read' | 'write';
+}
+
+export const LagPopover = ({text, type}: LagPopoverProps) => (
```
Raubzeug, line 14: component name is a bit confusing: this is not a popover actually.

src/components/LabelWithPopover/LabelWithPopover.tsx
```diff
@@ -3,18 +3,14 @@ import type {ReactNode} from 'react';
 import {HelpPopover} from '@gravity-ui/uikit';
 
 interface LabelWithPopoverProps {
-    headerText: string;
-    popoverContent: ReactNode;
+    text: string;
+    content: ReactNode;
```
Raubzeug, line 7: why did you decide to rename this property? As for me, previous name explains what property it is better.

src/components/LagPopover/LagPopover.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <div className={b({type})}>
+        <div>{text}</div>
```
Raubzeug, line 16: should we consider if text might overflow container?

Raubzeug (changes requested):

src/components/LabelWithPopover/LabelWithPopover.tsx
artemmufazalov: I didn't like `headerText` name as it's not a header. But I agree with you regarding `popoverContent` vs just `content`, the first one is better

src/components/LagPopover/LagPopover.tsx
artemmufazalov: I don't consider this component to be a universal popover content, it is used with topics, CDC streams and their sub entities. And there we have specific image with its short explanation (that is a little different for different entities). I don't think it's needed to consider overflow behaviour right away

src/components/InfoViewer/InfoViewer.tsx
Raubzeug: As far as we are not going to manipulate elements in this array, we may still use `infoIndex`, but more proper way is to add ids to `info` (maybe we can use `names` - as I remember, they are unique).

Raubzeug (dismissed):

Raubzeug (approved):

PR #340, opened 2023-03-24
fix(StorageNodes): make fqdn similar to nodes page

artemmufazalov:


src/utils/tooltip.js
```diff
@@ -109,34 +111,6 @@ const NodeTooltip = (props) => {
     );
 };
 
-const NodeEndpointsTooltip = (props) => {
```
artemmufazalov, line 112: I don’t like two things in these tooltips:
1. redux reducer as additional level - it complicates code
2. two different ways to display tooltips in the app - this redux tooltip and different components from uikit

So I prefer to use Popover, Popup and Tooltip from uikit, it makes it easier and more consistent. Therefore I move this components from here to /components/Tooltips

Raubzeug: I totally agree with you! Redux Tooltip is awful! Let's get rid of it step by step.

Raubzeug (changes requested):

src/containers/Storage/StorageNodes/StorageNodes.scss
```diff
@@ -24,13 +24,10 @@
...
-        display: inline-block;
-        overflow: hidden;
-
-        max-width: 330px;
+        display: flex;
+        align-items: center;
 
-        vertical-align: top;
-        text-overflow: ellipsis;
+        max-width: 305px;
```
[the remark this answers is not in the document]
artemmufazalov: In this case overflow behaviour is defined in `EntityStatus` component, so we don't need to define it additionally

src/containers/Nodes/getNodesColumns.tsx
```diff
@@ -47,27 +48,29 @@ export function getNodesColumns({
...
-                                className={b('external-button')}
-                                target="_blank"
-                            >
-                                <Icon name="external" />
-                            </Button>
-                        )}
+                        <Popover
+                            content={<NodeEndpointsTooltip data={row} />}
+                            placement={['top', 'bottom']}
+                            behavior={PopoverBehavior.Immediate}
```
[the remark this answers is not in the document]
Raubzeug: Ok!

src/components/Tooltips/NodeEndpointsTooltip/index.ts
```diff
@@ -0,0 +1 @@
+export * from './NodeEndpointsTooltip';
```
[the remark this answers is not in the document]
Raubzeug: Commonly we don't use such technique in other projects. It's useful for libs when you collect imports from a few files. But here we save a few letters when import a component and have a pile of `index` files.

src/components/Icon/Icon.tsx
```diff
@@ -31,3 +31,9 @@ export const Icon = ({
         />
     );
 };
+
+// When used with uikit components Icon is considered as text element and corresponding styles are applied
+// IconWrapper overrides displayName to 'Icon', so it will be considered as an icon with right styles
+export const IconWrapper = (props: IconProps) => <Icon {...props} />;
+
+IconWrapper.displayName = 'Icon';
```
Raubzeug, line 39: I realised that we even do not need Wrapper.
We may just do `Icon.displayName = 'Icon'` and that's it!

Raubzeug (changes requested):

src/components/Icon/Icon.tsx
Raubzeug: Ok, lets do an issue and refactor it in the future!

Raubzeug (approved):

src/components/Icon/Icon.tsx
artemmufazalov: Added issue - https://github.com/ydb-platform/ydb-embedded-ui/issues/344

PR #342, opened 2023-03-27
fix(TopicInfo): do not display /s when speed is undefined

artemmufazalov:


src/components/InfoViewer/formatters/pqGroup.ts
```diff
@@ -42,7 +42,7 @@ export const formatPQTabletConfig = createInfoFormatter<TPQTabletConfig>({
 export const formatPQPartitionConfig = createInfoFormatter<TPQPartitionConfig>({
     values: {
         StorageLimitBytes: formatBytes,
-        WriteSpeedInBytesPerSecond: formatBps,
+        WriteSpeedInBytesPerSecond: (value) => value && formatBps(value),
```
Raubzeug, line 45: maybe it is possible to check if not value in `formatBps`?

Raubzeug (changes requested):

src/components/InfoViewer/formatters/pqGroup.ts
artemmufazalov: Yep, added check there

Raubzeug (approved):

PR #357, opened 2023-04-17
fix(Storage): fix incorrect usage on zero available space

artemmufazalov:


src/store/reducers/storage.js
```diff
@@ -243,13 +243,11 @@ export const getFlatListStorageGroups = createSelector([getStoragePools], (stora
...
-                                (Number(vDisk.AvailableSize) ||
-                                    Number(vDisk.PDisk?.AvailableSize) ||
-                                    0) +
-                                (Number(vDisk.AllocatedSize) || 0)
-                            );
+                            const {AvailableSize, AllocatedSize, PDisk} = vDisk;
+                            const available = AvailableSize || PDisk?.AvailableSize || 0;
+                            const allocated = AllocatedSize || 0;
+
+                            return acc + Number(available) + Number(allocated);
```
artemmufazalov, line 250: If `AvailableSize` was `'0'`, `PDIsks.AvailableSize` was used, that is incorrect. Now we firstly check if it is defined (it is of `string` type), and only after convert it to `Number`

Raubzeug (approved):

PR #358, opened 2023-04-18
fix(TopShards): fix table crash on undefined values

artemmufazalov:


src/routes.ts
```diff
@@ -28,7 +28,7 @@ export const CLUSTER_PAGES = {
 
 export function createHref(
     route: string,
-    params?: object,
+    params?: Record<string, string | number>,
```
artemmufazalov, line 31: Previously it was possible to pass undefined values, though they caused error

src/containers/Tenant/Diagnostics/TopShards/TopShards.tsx
```diff
@@ -190,53 +191,60 @@ export const TopShards = ({tenantPath, type}: TopShardsProps) => {
             };
         };
 
-        const columns: Column<any>[] = [
+        const columns: Column<KeyValueRow>[] = [
             {
                 name: tableColumnsNames.Path,
-                render: ({value: relativeNodePath}) => {
+                render: ({row}) => {
```
artemmufazalov, line 197: Changed `value` to `row`, because `value` is of `unknown` type while `row` is typed with an argument to `Column` (`KeyValueRow` in this case)

src/containers/Tenant/Diagnostics/TopShards/TopShards.tsx
```diff
@@ -255,12 +263,22 @@ export const TopShards = ({tenantPath, type}: TopShardsProps) => {
             // after NodeId
             columns.splice(5, 0, {
                 name: tableColumnsNames.PeakTime,
-                render: ({value}) => formatDateTime(new Date(value as string).valueOf()),
+                render: ({row}) => {
+                    if (!row.PeakTime) {
```
Raubzeug, line 267: Lets move this into helper?
```
function renderDateTime(value?: string) {
    if (!value) {.....}
}
```

Raubzeug (changes requested):

Raubzeug (dismissed):

Raubzeug (approved):

PR #361, opened 2023-04-18
feat(QueryEditor): enable explain query modes

artemmufazalov:
Connected PRs:
- #360
- #362
- #363 
- #364 
- #367 
- #369

### DIFF
--- a/src/containers/Tenant/QueryEditor/QueryEditor.js
+++ b/src/containers/Tenant/QueryEditor/QueryEditor.js
@@ -4,41 +4,36 @@ import {connect} from 'react-redux';
 import cn from 'bem-cn-lite';
 import _ from 'lodash';
 import MonacoEditor from 'react-monaco-editor';
-import {Button, DropdownMenu} from '@gravity-ui/uikit';
 
 import SplitPane from '../../../components/SplitPane';
 import {QueryResultTable} from '../../../components/QueryResultTable';
 
-import SaveQuery from './SaveQuery/SaveQuery';
 import SavedQueries from './SavedQueries/SavedQueries';
-import {Icon} from '../../../components/Icon';
 import QueryResult from './QueryResult/QueryResult';
 import QueryExplain from './QueryExplain/QueryExplain';
+import {QueryEditorControls} from './QueryEditorControls/QueryEditorControls';
+import {OldQueryEditorControls} from './QueryEditorControls/OldQueryEditorControls';
 
 import {
-    sendQuery,
+    sendExecuteQuery,
     changeUserInput,
     saveQueryToHistory,
     goToPreviousQuery,
     goToNextQuery,
-    selectRunAction,
-    RUN_ACTIONS_VALUES,
     MONACO_HOT_KEY_ACTIONS,
     setMonacoHotKey,
 } from '../../../store/reducers/executeQuery';
 import {getExplainQuery, getExplainQueryAst} from '../../../store/reducers/explainQuery';
-import {getSettingValue, setSettingValue} from '../../../store/reducers/settings';
+import {getParsedSettingValue, setSettingValue} from '../../../store/reducers/settings';
 import {
     DEFAULT_IS_QUERY_RESULT_COLLAPSED,
     DEFAULT_SIZE_RESULT_PANE_KEY,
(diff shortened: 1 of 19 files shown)


src/containers/Tenant/QueryEditor/QueryExplain/QueryExplain.js
```diff
@@ -102,12 +102,6 @@ function QueryExplain(props) {
         };
     }, []);
 
-    useEffect(() => {
```
artemmufazalov, line 105: It occurred, that `explain-ast` is run in data-query mode, so when we request `ast` for `explain-scan` or `explain-query`, we have `ast` that does not correspond to initial mode. Thus, it should be displayed only if query response have it

src/containers/Tenant/QueryEditor/QueryEditorControls/QueryEditorControls.scss
```diff
@@ -0,0 +1,57 @@
+.ydb-query-editor-controls {
```
artemmufazalov, line 1: These styles was moved from `QueryEditor.scss` with minor changess (e.g. `&__mode-selector` was added)

src/containers/Tenant/QueryEditor/QueryEditor.js
```diff
@@ -4,25 +4,21 @@ import {connect} from 'react-redux';
...
-import SaveQuery from './SaveQuery/SaveQuery';
 import SavedQueries from './SavedQueries/SavedQueries';
-import {Icon} from '../../../components/Icon';
 import QueryResult from './QueryResult/QueryResult';
 import QueryExplain from './QueryExplain/QueryExplain';
+import {QueryEditorControls} from './QueryEditorControls/QueryEditorControls';
 
 import {
-    sendQuery,
+    sendExecuteQuery,
```
artemmufazalov, line 17: It was just renamed

src/store/reducers/executeQuery.ts
```diff
@@ -55,7 +46,6 @@ const initialState = {
                 ? MAXIMUM_QUERIES_IN_HISTORY - 1
                 : queriesHistoryInitial.length - 1,
     },
-    runAction: readSavedSettingsValue(QUERY_INITIAL_RUN_ACTION_KEY, RUN_ACTIONS_VALUES.script),
```
artemmufazalov, line 58: Selected query mode is now managed inside `QueryEditor` component

Raubzeug:
When the new mode is enabled, it is unclear, what action is produced by the button `Type: Script | Scan` (actually, no action). I would better make this button similar with query type selection (in the old mode).

0EEB6DB9-FB6B-4BDE-B3E1-871F644EE16D

Raubzeug (changes requested):

src/containers/Tenant/QueryEditor/QueryEditor.js
```diff
@@ -633,22 +593,25 @@ const mapStateToProps = (state) => {
         executeQuery: state.executeQuery,
         explainQuery: state.explainQuery,
         savedQueries: parseJson(getSettingValue(state, SAVED_QUERIES_KEY)),
```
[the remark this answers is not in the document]
artemmufazalov: We don't need to parse value if it used as a string value, so `getSettingValue` doesn't always go with `parseJson` . As for me, adding some additional `getParsedSettingValue` helper will not make code easier as it won't be clear when it should be used and when we should use just `getSettingValue`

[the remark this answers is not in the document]
Raubzeug: I suggested having two helpers! But it's totally up to you!
Unfortunately now it is not clear too - when we should parse JSON, and when value is a plain string... What if add `try ... catch`? If `JSON.parse` fails, return unparsed value, and if succeed - return parsed.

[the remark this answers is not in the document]
artemmufazalov: Agree with you, now it's unclear too. Added additional parser, maybe I'll replace old one with it when I'll be refactoring

Raubzeug (approved):

Raubzeug (approved):

PR #362, opened 2023-04-19
fix(QueryExplain): request AST if it is empty

artemmufazalov:
`explain-ast` request is sent every time we open AST tab for explain response with lacking `ast`. But if we open AST and then run explain, which returns no `ast`, the tab will be empty and no additional request for `explain-ast` will be sent. This PR fixes it

Part of #361

### DIFF
--- a/src/containers/Tenant/QueryEditor/QueryExplain/QueryExplain.js
+++ b/src/containers/Tenant/QueryEditor/QueryExplain/QueryExplain.js
@@ -116,7 +116,7 @@ function QueryExplain(props) {
         if (!props.ast && activeOption === ExplainOptionIds.ast) {
             props.astQuery();
         }
-    }, [activeOption]);
+    }, [activeOption, props.ast]);
 
     const onSelectOption = (tabId) => {
         setActiveOption(tabId);


src/containers/Tenant/QueryEditor/QueryExplain/QueryExplain.js
```diff
@@ -116,7 +116,7 @@ function QueryExplain(props) {
         if (!props.ast && activeOption === ExplainOptionIds.ast) {
             props.astQuery();
         }
-    }, [activeOption]);
+    }, [activeOption, props.ast]);
```
Raubzeug, line 119: I'd probably move `renderAstExplain` to the separate component.
And this `useEffect` seems to be redundant. You may simply check if-statement in main flow.

Raubzeug (changes requested):

src/containers/Tenant/QueryEditor/QueryExplain/QueryExplain.js
artemmufazalov: `QueryExplain` should be split to several components and migrated to TS, but it isn't the issue of this PR.

As for the `useEffect`, I don't agree, without `useEffect` request will be sent at least twice as request toggles component loading state which makes component to rerender. Also it could be bugged in case of unsuccessful requests. Request fail -> component rerender because of a new error -> request is sent again -> loading state toggled -> new request is sent, infinite loader is shown instead of an error

Raubzeug: We may get rid of `useEffect` when move Ast to new component: just `error` should be added into if-statement.
Loader is in parent component, so when it is shown, `<Ast/>` would not be rendered. When loading is done, `props.ast` or `props.error` is here.

artemmufazalov: I tried separating `AST`. Without hooks it still causes infinite data fetches. I don't know how to ensure single fetch without wrapping either fetch or component itself in a hook. I'd prefer fetch data in hook and separate `AST` with complex `QueryExplain` component refactoring

Raubzeug: Ok, let's think about it when refactor `QueryExplain`.

Raubzeug (approved):

Raubzeug (approved):

PR #365, opened 2023-04-19
refactor(QueriesHistory): use useTypedSelector hook

artemmufazalov:


src/containers/Tenant/QueryEditor/QueriesHistory/QueriesHistory.tsx
```diff
@@ -17,8 +18,7 @@ interface QueriesHistoryProps {
 
 function QueriesHistory(props: QueriesHistoryProps) {
     const [isHistoryVisible, setIsHistoryVisible] = useState(false);
-    const history: string[] =
-        useSelector((state: any) => state.executeQuery.history?.queries) ?? [];
+    const history = useTypedSelector((state) => state.executeQuery.history.queries) ?? [];
```
Raubzeug, line 21: we don't need to check if `history` exists in state?

Raubzeug (approved):

src/containers/Tenant/QueryEditor/QueriesHistory/QueriesHistory.tsx
artemmufazalov: We update history in actions, that are all typed now. And `executeQuery` reducer state is typed too. I hope there is no need in the additional check and typecheck will tell us if anything will be broken

PR #367, opened 2023-04-21
fix(query): correctly process NetworkError on actions failure

artemmufazalov:


src/store/reducers/executeQuery.ts
```diff
@@ -82,6 +83,14 @@ const executeQuery: Reducer<ExecuteQueryState, ExecuteQueryAction> = (
         }
         // 401 Unauthorized error is handled by GenericAPI
         case SEND_QUERY.FAILURE: {
+            if (isNetworkError(action.error)) {
```
Raubzeug, line 86: Maybe it should be any `ParseRequestError` common function? In 3 places parsing logic is the same.

Raubzeug (changes requested):

Raubzeug (dismissed):

Raubzeug (approved):

PR #382, opened 2023-05-11
feat(Partitions): display partitions for topic without consumers

artemmufazalov:
- Replace `getNodes` with `getNodesList` for better performance
- Use selectors for `consumers` and `nodesMap`
- Remove `PartitionsWrapper` (data fetch logic changed, it no longer fits)
- Separate `PartitionsControls` component
- Use `hiddenColumns` instead of `selectedColumns` list, since it better fits for table with different set of columns for different entities
- Enable consumer select to be without consumer. Fetch partitions data from `/describe_topic` for this case


src/types/store/consumer.ts
```diff
@@ -11,32 +11,32 @@ import type {DescribeConsumerResult} from '../api/consumer';
...
 
 // All fields should be present though they could be undefined
-export interface IPreparedPartitionData {
+export interface PreparedPartitionData {
     partitionId: string;
     storeSize: string;
 
     writeSpeed: IProcessSpeedStats;
-    readSpeed: IProcessSpeedStats;
+    readSpeed: IProcessSpeedStats | undefined;
```
artemmufazalov, line 19: Fields, that could be `undefined` are empty for partitions without consumers

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    const handlePartitionIdSearchChange = (value: string) => {
+        setPartitionIdSearchValue(value);
+    };
+
+    const handleGeneralSearchChange = (value: string) => {
+        setGeneralSearchValue(value);
+    };
+
+    const hadleTableColumnsSetupChange = (value: TableColumnSetupItem[]) => {
```
artemmufazalov, line 116: Hidden columns are used now, because set of columns differs for partitions with consumers and partitions without consumers

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const result = [...hiddenColumns];
```
artemmufazalov, line 117: Hidden columns are used now, because set of columns differs for partitions with consumers and partitions without consumers

src/containers/Tenant/Diagnostics/Partitions/utils/index.ts
```diff
@@ -0,0 +1,26 @@
+import type {PreparedPartitionData} from '../../../../../types/store/consumer';
+import type {NodesMap} from '../../../../../types/store/nodesList';
+import type {PreparedPartitionDataWithHosts} from './types';
+
+export const addHostToPartitions = (
```
artemmufazalov, line 5: It was moved from `Partitions` to reduce component size

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
```diff
@@ -43,215 +40,162 @@ export const b = block('ydb-diagnostics-partitions');
...
-
+export const Partitions = ({path}: PartitionsProps) => {
     const dispatch = useDispatch();
 
-    const [generalSearchValue, setGeneralSearchValue] = useState('');
-    const [partitionIdSearchValue, setPartitionIdSearchValue] = useState('');
-
+    // Manual path control to ensure it updates with state reset
+    // so no request with wrong params will be sent
     const [componentCurrentPath, setComponentCurrentPath] = useState(path);
```
Raubzeug, line 50: When new `path` arrives, it triggers component's re-render. What do we need to ensure?

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
```diff
@@ -43,215 +40,162 @@ export const b = block('ydb-diagnostics-partitions');
...
-                    items={columnsToSelect}
-                    showStatus
-                    onUpdate={hadleTableColumnsSetupChange}
-                    className={b('table-settings')}
-                />
-            </div>
+            <PartitionsControls
+                consumers={consumers}
+                selectedConsumer={selectedConsumer}
+                handleSelectedConsumerChange={handleSelectedConsumerChange}
```
Raubzeug, line 189: Just to reduce ambiguity - props should be called `onSomething`, its value - `handleSomething`

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
```diff
@@ -43,215 +40,162 @@ export const b = block('ydb-diagnostics-partitions');
...
-                title: PARTITIONS_COLUMNS_TITILES[id],
-                selected: Boolean(selectedColumns?.includes(id)),
-                id: id,
-                required: id === PARTITIONS_COLUMNS_IDS.PARTITION_ID,
-            };
-        });
-    }, [selectedColumns]);
+    }, [dispatch, topicWasLoaded, selectedConsumer, consumers]);
 
     const columnsToShow = useMemo(() => {
```
Raubzeug, line 146: This may be moved into custom hook `useVisibleColumns = (initialColumns, hiddenColumns) => {return [visibleColumns, setColumns]}`
`hiddenColumns` - may be even key for LS, if it is common to store hidden columns for different tables in LS.

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
```diff
@@ -43,215 +40,162 @@ export const b = block('ydb-diagnostics-partitions');
...
+        dispatch(setTopicDataWasNotLoaded());
+
+        dispatch(getNodesList());
+        dispatch(getTopic(path));
+
         setComponentCurrentPath(path);
     }, [dispatch, path]);
 
-    const fetchConsumerData = useCallback(
+    // Data, its source and presentation is different for partitions with and without consumers
```
Raubzeug, line 99: is it possible to split logic into two custom hooks - handle data with and without consumers?

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
```diff
@@ -0,0 +1,175 @@
...
+                filterable={consumers && consumers.length > 5}
+                disabled={selectIsDisabled || !consumers || !consumers.length}
+                open={consumerSelectOpen}
+                onOpenChange={setConsumerSelectOpen}
+                // Although only one value could be selected
+                // Multiple type Select is used
+                // The reason - we need Select to be able to work with no value
+                // And it is easier to make multiple Select close on value change
+                // Than to enable single select to work with empty values
+                multiple
```
Raubzeug, line 151: For me it is not the best way to do it, it looks like a crutch. The better choice is to make a wrapper for `Select` that will allow to undo selection (the common way is - when selected item in Select popup is clicked again, selections cancels).

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
```diff
@@ -0,0 +1,175 @@
...
+                readerName,
+                readSessionId,
+                partitionNodeId,
+                connectionNodeId,
+                partitionHost,
+                connectionHost,
+            } = partition;
+
+            const isPartitionIdMatch = partitionIdRe.test(partitionId);
+            const isOtherValuesMatch =
```
Raubzeug, line 65:
```
const otherValues = [readerName, readSessionId, ...].filter(Boolean).map(String)`
const isOtherValuesMatch = otherValues.some(...)
```

src/store/reducers/topic.ts
```diff
@@ -142,4 +151,53 @@ export const selectPreparedConsumersData: Selector<
...
+            consumerReadIdleTime: undefined,
+
+            uncommitedMessages: undefined,
+            unreadMessages: undefined,
+
+            startOffset: start_offset,
+            endOffset: end_offset,
+            commitedOffset: undefined,
+
+            readSessionId: undefined,
```
Raubzeug, line 194: Could you please clarify if these properties may be not undefined?

Raubzeug (changes requested):

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
artemmufazalov: Single select with ability to undo selection looks like a crutch as well, but it more difficult in terms of realisation. That's why I wrote this comment. Multiple Select crutch is much easier

src/store/reducers/topic.ts
artemmufazalov: They are always `undefined` for partitions from `/describe_topic` endpoint. And always defined for `/describe_consumer`

src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
artemmufazalov: https://github.com/ydb-platform/ydb-embedded-ui/issues/386

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
artemmufazalov: As for it won’t make it easier, as there is a problem - there still should be some logic that will determine what entity we are dealing with

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
artemmufazalov: I’m not sure that there is much use in moving columns into a custom hook. However, I created a hook `useSetting` to simplify interaction with stored settings

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
artemmufazalov: No, this way it won’t render correctly on autorefresh. Also first render could happen with already loaded topic data (when navigating from info or consumers tabs)

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
Raubzeug: Yes, but now this logic is splitted between useEffect and fetchData. My point was that you pass `selectedConsumer` into the hook and it returns `fetchData, columns, columnsIds` to isolate logic connected with difference you mentioned.

src/utils/hooks/useSetting.ts
```diff
@@ -0,0 +1,23 @@
...
+import {useTypedSelector} from './useTypedSelector';
+
+export const useSetting = <T>(key: string, defaultValue: T): [T, (value: T) => void] => {
+    const dispatch = useDispatch();
+
+    const settingValue: T = useTypedSelector(
+        (state) => getParsedSettingValue(state, key) || defaultValue,
+    );
+
+    const setValue = useCallback(
```
Raubzeug, line 15: why did you decide to use `useCallback` here?

artemmufazalov: I decided to add `useCallback` to ensure, that return values are properly memoized and could be safely used as dependencies in other hooks

src/containers/Tenant/Diagnostics/Partitions/Partitions.tsx
```diff
@@ -152,25 +130,26 @@ export const Partitions = ({path}: PartitionsProps) => {
...
+    }
 
     const getContent = () => {
-        if (loading) {
+        if (partitionsLoading && !partitionsWasLoaded) {
             return <TableSkeleton className={b('loader')} />;
         }
-        if (error) {
-            return <ResponseError error={error} />;
+        if (partitionsError) {
```
Raubzeug, line 145: Why did you move this error to `getContent`?

Raubzeug (changes requested):

Raubzeug (approved):

Raubzeug (approved):

PR #395, opened 2023-05-24
Fix partitions tab

artemmufazalov:
`Select` with ability to unselect by clicking on the selected option occurred to be not clear for the users. Added explicit "No consumer" option to `Select`

Screen Shot 2023-05-24 at 13 19 20


src/containers/Tenant/Diagnostics/Partitions/PartitionsControls/PartitionsControls.tsx
```diff
@@ -137,25 +132,24 @@ export const PartitionsControls = ({
...
+        return (
+            <div className={b('select-option', {empty: option.value === ''})}>{option.content}</div>
+        );
+    };
+
     return (
         <div className={b('controls')}>
             <Select
                 className={b('consumer-select')}
-                placeholder={i18n('controls.consumerSelector.placeholder')}
```
artemmufazalov, line 144: We don't need placeholder, as we now always have empty option ({value: "", content: "No consumer"})

Raubzeug (approved):

PR #396, opened 2023-05-24
fix(UserSettings): separate Setting, enable additional settings

artemmufazalov:


src/containers/UserSettings/UserSettings.tsx
```diff
@@ -1,140 +1,104 @@
...
-                popoverContent={
-                    'Use /viewer/json/nodes endpoint for Nodes Tab in diagnostics. It returns incorrect data on older versions'
-                }
-            />
-        );
-    };
+const themeValues = [
+    {
+        value: Theme.system,
+        content: 'System',
```
Raubzeug, line 30: i18n?

src/containers/UserSettings/Setting.tsx
```diff
@@ -0,0 +1,84 @@
...
+    helpPopoverContent?: ReactNode;
+    values?: {value: string; content: string}[];
+    defaultValue?: unknown;
+}
+
+export const Setting = ({
+    type = 'switch',
+    settingKey,
+    title,
+    withHelpPopover,
```
Raubzeug, line 28: It seems to be redundant. You may just check if `helpPopoverContent` is presented.

Raubzeug (changes requested):

Raubzeug (approved):

PR #400, opened 2023-05-25
refactor: migrate settings reducer to ts

artemmufazalov:


src/store/reducers/settings/types.ts
```diff
@@ -0,0 +1,23 @@
+import {changeFilter, SET_SETTING_VALUE} from './settings';
+
+export enum ProblemFilterValues {
```
Raubzeug, line 3: Lets try not to use enums (https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums)

Raubzeug (changes requested):

Raubzeug (approved):

PR #401, opened 2023-05-25
refactor(Tablet): add and use ContentWithPopup

artemmufazalov:


src/components/Tablet/Tablet.tsx
```diff
@@ -1,59 +1,35 @@
...
-
+export const Tablet = ({tablet = {}}: TabletProps) => {
     const {TabletId: id} = tablet;
     const status = tablet.Overall?.toLowerCase();
 
     return (
-        <InternalLink
-            onClick={_onTabletClick}
-            to={id && createHref(routes.tablet, {id})}
+        <WithPopupWrapper
```
Raubzeug, line 24: It seems you wanted to make a HOC... and I think HOC would be better in this case.

Raubzeug (changes requested):

src/components/Tablet/Tablet.tsx
artemmufazalov: The problem with HOC is that there is a need to pass props from the wrapped component to the TooltipContent, which will be a param for that HOC

Raubzeug (approved):

PR #404, opened 2023-05-26
refactor(UserSettings): separate settings

artemmufazalov:


src/containers/UserSettings/settings.ts
```diff
@@ -0,0 +1,92 @@
...
+    title: i18n('settings.useNodesEndpoint.title'),
+    helpPopoverContent: i18n('settings.useNodesEndpoint.popover'),
+};
+const enableQueryModesForExplainSetting: SettingProps = {
+    settingKey: ENABLE_QUERY_MODES_FOR_EXPLAIN,
+    title: i18n('settings.enableQueryModesForExplain.title'),
+    helpPopoverContent: i18n('settings.enableQueryModesForExplain.popover'),
+};
+
+export const settings: YDBEmbeddedUISettings = {
```
Raubzeug, line 65: If string which may be treated will be used as key in this objects, order of rendering may be changed surprisingly. As for me it's more robust to use arrays.

Raubzeug (changes requested):

Raubzeug (approved):

PR #405, opened 2023-05-30
fix(Tablet): fetch data on action finish

artemmufazalov:


Raubzeug (changes requested):

src/containers/Tablet/TabletControls/TabletControls.tsx
```diff
@@ -17,18 +17,19 @@ type VisibleDialogType = EVisibleDialogType | null;
...
 }
 
-export const TabletControls = ({tablet}: TabletControlsProps) => {
+export const TabletControls = ({tablet, fetchData}: TabletControlsProps) => {
     const {TabletId, HiveId} = tablet;
 
     const [isDialogVisible, setIsDialogVisible] = useState(false);
     const [visibleDialogType, setVisibleDialogType] = useState<VisibleDialogType>(null);
-    const [isTabletActionsDisabled, setIsTabletActionsDisabled] = useState(false);
+    const [isTabletActionLoading, setIsTabletActionLoading] = useState(false);
```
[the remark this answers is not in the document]
artemmufazalov: Here we control state internally, because it is affected only by actions, triggered by `TabletControls`, that are defined in `TabletControls` as well.
For regular fetches (external) loading state should not be set (still it resets to true when new data is received) . 

Probably it all can be changed with component logic redesign, but I don’t want to do it for the current issue

src/containers/Tablet/TabletControls/TabletControls.tsx
```diff
@@ -17,18 +17,19 @@ type VisibleDialogType = EVisibleDialogType | null;
...
 
     // Enable controls after data update
     useEffect(() => {
-        setIsTabletActionsDisabled(false);
+        setIsTabletActionLoading(false);
```
Raubzeug, line 32: instead of this `useEffect` lets make a function
```
const onConfirmActionFinish = () => {
  fetchData()
  setIsTabletActionLoading(false)
}
```

src/containers/Tablet/TabletControls/TabletControls.tsx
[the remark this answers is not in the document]
Raubzeug: Ok, let's stick to this, but remove reseting when new data is received (see next comment).

Raubzeug (changes requested):

src/containers/Tablet/TabletControls/TabletControls.tsx
artemmufazalov: In this case buttons loading state resets before data is loaded (tablet state updated). So it will be possible to click "stop" twice

Raubzeug: To handle this you may pass `tabletDataIsLoading` and set `loading = isTabletActionLoading || tabletDataIsLoading`

artemmufazalov: Now we have different loading state for tablet content and for the buttons.

Buttons loading state is needed to prevent unavailable actions. The thing is that actions is connected with tablet `State` and for different states different set of actions is available. When you trigger an action, tablet `State` could change, so to enable buttons we await actual tablet data with the latests `State`.

Additionally, we have tablet data autoupdate, which happens every 30 seconds. If to pass this state to tablet controls, they will be unavailable every 30 seconds as well. It’s not correct, because it isn’t caused by action that could trigger tablet `State` change.

To sum up:
1) Buttons loading state should be connected with tablet `State` actuality
2) We need actual tablet `State`, but cannot use tablet `loading` to await, because it triggers loading when it isn't needed
3) We cannot reset buttons state in `confirmAction`, because this way we don't have actual `State` when reset

Thus we have two different states with different logic.

Raubzeug (approved):

PR #407, opened 2023-05-31
fix(ClusterInfo): update versions bar, rework DC and Tablets fields

artemmufazalov:
- Move DC and Tablets from title to info section
- Update Internal Viewer and additional links view
- Use versions progress as `InfoViewer` value, display versions list below, color versions in the list

## Before
Screen Shot 2023-05-31 at 15 34 55

## After
Screen Shot 2023-05-31 at 15 35 10


src/containers/Header/Header.tsx
```diff
@@ -2,17 +2,17 @@ import React, {useEffect} from 'react';
...
 import Divider from '../../components/Divider/Divider';
-import {Icon} from '../../components/Icon';
 
 import {backend, customBackend} from '../../store';
 import {getHostInfo} from '../../store/reducers/host';
 import {HeaderItemType} from '../../store/reducers/header';
 import {useTypedSelector} from '../../utils/hooks';
 
 import './Header.scss';
+import {ExternalLinkWithIcon} from '../../components/ExternalLinkWithIcon/ExternalLinkWithIcon';
```
Raubzeug, line 15: imports order

src/containers/Cluster/ClusterInfo/ClusterInfo.tsx
```diff
@@ -158,36 +186,30 @@ export const ClusterInfo = ({
         return <ResponseError error={error || nodesError} />;
     }
 
+    let internalLink = backend + '/internal';
+
+    if (singleClusterMode && !customBackend) {
+        internalLink = `/internal`;
+    }
+
+    const {info = [], links = []} = additionalClusterProps;
```
Raubzeug, line 195: Inside `getInfo` there are also fallbacks. It seems to be enough callbacks inside the function.

artemmufazalov: I use spread operator on `links` below. So it's needed here as well. Removed fallbacks from `getInfo`

Raubzeug (approved):

PR #410, opened 2023-06-01
chore: add unimported command, delete some unused files

artemmufazalov:
For now only a command added. Once the command will be used in CI, custom config should be added as well.

Also deleted some unused files (but not all).


Raubzeug (approved):

Raubzeug (changes requested):
Did you add a config for `unimported`?

Raubzeug (approved):

PR #418, opened 2023-06-07
feat: rework navigation, update breadcrumbs

artemmufazalov:
Breadcrumbs are not ready yet, it's an intermediary version

### DIFF
--- a/src/containers/AsideNavigation/AsideNavigation.tsx
+++ b/src/containers/AsideNavigation/AsideNavigation.tsx
@@ -9,11 +9,7 @@ import {AsideHeader, MenuItem as AsideHeaderMenuItem, FooterItem} from '@gravity
 
 import signOutIcon from '../../assets/icons/signOut.svg';
 import signInIcon from '../../assets/icons/signIn.svg';
-import databaseIcon from '../../assets/icons/server.svg';
-import storageIcon from '../../assets/icons/storage.svg';
-import clusterIcon from '../../assets/icons/cluster.svg';
 import ydbLogoIcon from '../../assets/icons/ydb.svg';
-import databasesIcon from '../../assets/icons/databases.svg';
 import userSecret from '../../assets/icons/user-secret.svg';
 import userChecked from '../../assets/icons/user-check.svg';
 import settingsIcon from '../../assets/icons/settings.svg';
@@ -21,7 +17,7 @@ import supportIcon from '../../assets/icons/support.svg';
 
 import {UserSettings} from '../UserSettings/UserSettings';
 
-import routes, {createHref, CLUSTER_PAGES} from '../../routes';
+import routes, {createHref} from '../../routes';
 
 import {logout} from '../../store/reducers/authentication';
 import {getParsedSettingValue, setSettingValue} from '../../store/reducers/settings/settings';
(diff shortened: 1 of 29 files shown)


src/containers/AsideNavigation/AsideNavigation.tsx
```diff
@@ -115,46 +111,8 @@ interface AsideNavigationProps {
...
-        icon: clusterIcon,
-        iconSize: 20,
-        location: createHref(routes.cluster, {
-            activeTab: CLUSTER_PAGES.cluster.id,
-        }),
-        locationKeys: ['/cluster/cluster'],
-    },
-];
+// FIXME: add items or delete
+const items: MenuItem[] = [];
```
artemmufazalov, line 115: I left it as it was to keep logic and to fill it with new items on the next iteration

src/containers/Tenants/Tenants.js
```diff
@@ -94,14 +86,15 @@ class Tenants extends React.Component {
...
             <div className={b('controls')}>
-                <TextInput
-                    className={b('search')}
-                    placeholder="Database name"
-                    value={searchQuery}
-                    onUpdate={handleSearchQuery}
-                    hasClear
-                    autoFocus
-                />
+                <div className={b('search')}>
```
artemmufazalov, line 89: `TextInput` classname sometimes incorrectly applied, so I wrapped it with a `div` with `classname` instead

src/containers/Versions/Versions.tsx
```diff
@@ -55,67 +64,70 @@ export const Versions = ({nodes = [], versionToColor}: VersionsProps) => {
...
-    const renderGroupedNodes = () => {
-        const tenantNodes = getGroupedTenantNodes(nodes, versionToColor, groupByValue);
-        const storageNodes = getGroupedStorageNodes(nodes, versionToColor);
-        const otherNodes = getOtherNodes(nodes, versionToColor);
-        const storageNodesContent = storageNodes?.length ? (
-            <>
-                <h3>Storage nodes</h3>
-                {storageNodes.map(({title, nodes: itemNodes, items, versionColor}, index) => (
+
+    if (loading && !wasLoaded) {
```
artemmufazalov, line 68: Here I removed `renderGroupedNodes` function and added loader. GH considers its as a new code :(

src/containers/Cluster/ClusterInfo/ClusterInfo.tsx
```diff
@@ -204,20 +145,45 @@ export const ClusterInfo = ({
...
+        return (
+            <EntityStatus
+                size="m"
+                status={cluster?.Overall}
+                name={cluster?.Name ?? 'Unknown cluster'}
+                className={b('title')}
+            />
+        );
+    };
+
```
artemmufazalov, line 174: Two different skeletons for the info and for the title to include them in `ClusterInfo` layout and to enable expand button for the skeleton as well. This should enable user to hide content even if it's not loaded yet. May be helpful for big clusters, where `ClusterInfo` could load 20+ seconds, but there could be a need to to view tabs content ASAP

src/containers/Cluster/Cluster.tsx
```diff
@@ -1,59 +1,175 @@
...
+        <div className={b()}>
+            <ClusterInfo
+                cluster={cluster}
+                versionsValues={versionsValues}
+                loading={infoLoading}
+                error={clusterError}
+                additionalClusterProps={additionalClusterProps}
+            />
+
+            <div>
```
Raubzeug, line 148: What do you think about moving tabs into separate component?

Raubzeug (changes requested):

src/containers/Cluster/Cluster.tsx
artemmufazalov: I tried and not very liked it. It doesn't help much, but become complicated when we need to customise it (e.g. in Diagnostics there is autorefresh switch that is added to `Tabs`). So, in my view, it doesn't worth it

Raubzeug (approved):

src/containers/Versions/Versions.tsx
```diff
@@ -55,67 +64,70 @@ export const Versions = ({nodes = [], versionToColor}: VersionsProps) => {
...
+            )}
+        </>
+    ) : null;
+    const otherNodesContent = otherNodes?.length ? (
+        <>
+            <h3>Other nodes</h3>
+            {otherNodes.map(
+                ({title, nodes: itemNodes, items, versionColor, versionsValues}, index) => (
+                    <GroupedNodesTree
+                        key={`other-nodes-${index}`}
```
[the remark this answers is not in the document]
artemmufazalov: Changed to `title`. Titles are either DBs names, or versions, both should be unique in such view

PR #422, opened 2023-06-09
feat(QueryEditor): add data and query modes

artemmufazalov:


src/types/api/query.ts
artemmufazalov, line 1: Renamed `ScanPlan`, `ExplainScan`, `ExecuteScan` to `QueryPlan`, `ExplainQuery` and `ExecuteQuery` correspondingly as they correspond to `meta.type = "query"`

src/containers/Tenant/QueryEditor/QueryEditorControls/OldQueryEditorControls.tsx
```diff
@@ -21,12 +28,19 @@ export const OldQueryEditorControls = ({
     onUpdateQueryMode,
     queryMode,
 }: QueryEditorControlsProps) => {
+    // On setting change when 'data' or 'query' option selected
+    useEffect(() => {
```
Raubzeug, line 32: I don't think this logic should be here. Component should not be aware about "another" modes. As for me, `queryMode` should be calculated in selector.

Raubzeug (changes requested):

Raubzeug (approved):

Raubzeug (approved):

PR #425, opened 2023-06-15
fix(Tablet): clear tablet data on unmount

artemmufazalov:
Tablet describe is requested conditionally. For some tablets it should not be requested, but still requested because of old params, that satisfy condition. Old data could be displayed as well if not overwritten with a new entity. To fix this, old data is cleared on component unmount


src/containers/Tablet/Tablet.tsx
```diff
@@ -48,6 +48,12 @@ export const Tablet = () => {
         error,
     } = useTypedSelector((state) => state.tablet);
 
+    useEffect(() => {
```
Raubzeug, line 51: Does it work properly with StrictMode (double component mount)?

artemmufazalov: We are on react 17, there is no double mount in strict mode here

Raubzeug: It might be problems in this place when we decide to migrate to React 18. Lets mark it not to forget to pay attention to this piece of code.

Raubzeug (approved):

PR #427, opened 2023-06-16
feat(QueryEditor): transform history and saved to tabs

artemmufazalov:


src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -276,6 +279,7 @@ function ObjectSummary(props: ObjectSummaryProps) {
     const onOpenPreview = () => {
         dispatch(setShowPreview(true));
         dispatch(setTopLevelTab(TENANT_GENERAL_TABS_IDS.query));
+        dispatch(setQueryTab(TENANT_QUERY_TABS_ID.newQuery));
```
Raubzeug, line 282: Technically, table preview doesn't depend on the upper section content. Do you think it is required to force change tab to new query?

src/containers/Tenant/QueryEditor/SavedQueries/SavedQueries.tsx
```diff
@@ -0,0 +1,151 @@
...
+        {
+            name: 'body',
+            header: 'Query Text',
+            render: ({row: query}) => (
+                <div className={b('query')}>
+                    <div className={b('query-body')}>
+                        <TruncatedQuery value={query.body} maxQueryHeight={MAX_QUERY_HEIGHT} />
+                    </div>
+                    <span className={b('control-delete')} onClick={onDeleteQueryClick(query.name)}>
+                        <IconWrapper
```
Raubzeug, line 112: Icon looks pale and poorly visible. May be change it's color to primary and set size 16?

Raubzeug: As user I expect `edit` button as well. Because it's not clear that I will be able to edit my query after click the row. What do you think about adding it? And may be default action should be `edit`, not `save as new`

src/containers/Tenant/QueryEditor/SavedQueries/SavedQueries.tsx
```diff
@@ -0,0 +1,151 @@
...
+                </div>
+            ),
+            sortable: false,
+        },
+    ];
+
+    return (
+        <>
+            <div className={b()}>
+                <DataTable
```
Raubzeug, line 129: It seems that search by name or by query text would be very helpful here and in `History` as well.

Raubzeug (changes requested):

artemmufazalov:
> In design `Save` button is placed near `Explain`, not in the right corner.

There is no query type selector in this design. So it is not an actual versions of the controls

src/containers/Tenant/QueryEditor/SavedQueries/SavedQueries.tsx
artemmufazalov: We'll discuss it on call. It's not the issue of this PR

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
artemmufazalov: It's rendered in query result section. Though it doesn't depend on query editor content, it seems more suitable there, I don't thinks we should mess it with QueryTables

src/containers/Tenant/QueryEditor/SavedQueries/SavedQueries.tsx
artemmufazalov: Missed comment about buttons colors, I'll update them as well

artemmufazalov: Updated buttons colors as well

artemmufazalov: Where should be such a button? And do we really need it if we can use default button behaviour via option in dropdown menu?

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
Raubzeug: The case is: "I am watching queries history and I want to match it with particular table preview."
In the previous interface version it was possible. I think this should be discussed with PM, because I don't know frequency and importance of this case.

src/containers/Tenant/QueryEditor/SavedQueries/SavedQueries.tsx
Raubzeug: Yeah it seems it could be enough!

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
artemmufazalov: I still don't understand this case. For the table columns, their name and types there is schema under the tree. Preview is mostly for the table content. I don't see an issue where there is a need to match table content and particular queries within one tab. 

Additionally, discussed it with PM. It's seems, that this case isn't important

PR #430, opened 2023-06-20
fix(Cluster): add icons to tabs

artemmufazalov:
Screen Shot 2023-06-20 at 18 27 00


config-overrides.js
```diff
@@ -11,6 +12,7 @@ module.exports = {
             include: [
                 path.resolve(srcRoot, 'assets/icons'),
                 path.resolve(uiKitRoot, 'assets/icons'),
+                uiKitIconsRoot,
```
artemmufazalov, line 15: I don't know why it was resolved to 'assets/icons', so I left it as it was, but added path to uikit icons

Raubzeug (approved):

PR #431, opened 2023-06-21
feat(Tenant): transform general tabs into left navigation items

artemmufazalov:
Screen Shot 2023-06-21 at 11 58 20


Raubzeug:
1) It doesn't warn user about unsaved changes in query editor.
2) If I run query and thereafter change database, query text is still in editor, but query doesn't work. It seems that editor should be cleared upon change database.

src/containers/AsideNavigation/AsideNavigation.tsx
```diff
@@ -124,32 +126,65 @@ function AsideNavigation(props: AsideNavigationProps) {
 
     const [visiblePanel, setVisiblePanel] = useState<Panel>();
 
+    const [initiaTenantTab, setTenantInitialTab] = useSetting<string>(
```
Raubzeug, line 129: typo `initia`

src/containers/AsideNavigation/AsideNavigation.tsx
```diff
@@ -15,18 +18,20 @@ import userChecked from '../../assets/icons/user-check.svg';
 import settingsIcon from '../../assets/icons/settings.svg';
 import supportIcon from '../../assets/icons/support.svg';
 
-import {UserSettings} from '../UserSettings/UserSettings';
-
-import routes, {createHref} from '../../routes';
-
 import {logout} from '../../store/reducers/authentication';
 import {getParsedSettingValue, setSettingValue} from '../../store/reducers/settings/settings';
+import {TENANT_GENERAL_TABS_IDS} from '../../store/reducers/tenant/constants';
```
Raubzeug, line 23: Its no longer tabs. Maybe it should be renamed in separate PR or here in additional commit?

src/containers/AsideNavigation/AsideNavigation.tsx
```diff
@@ -124,32 +126,65 @@ function AsideNavigation(props: AsideNavigationProps) {
...
+    const {topLevelTab = initiaTenantTab} = useTypedSelector((state) => state.tenant);
+
     const setIsCompact = (compact: boolean) => {
         props.setSettingValue(ASIDE_HEADER_COMPACT_KEY, JSON.stringify(compact));
     };
 
+    const {pathname} = location;
+    const queryParams = parseQuery(location);
+
+    const isTenantPage = pathname === '/tenant';
```
Raubzeug, line 142: Lets start to move app paths into variables?

Raubzeug (changes requested):

artemmufazalov:
> 1. It doesn't warn user about unsaved changes in query editor.

It doesn't warn in two cases:
1) there is no such query in queries history (only last item is checked). So if you run query, then try to reload, it doesn't warn, yes. But this query will be the last in history
```
const hasUnsavedInput = input
            ? input !== history.queries[history.queries?.length - 1]
            : false;
```
2) if active window isn't `QueryEditor`. I'm not sure, that we should warn in such cases

Raubzeug:
> 1. there is no such query in queries history (only last item is checked). So if you run query, then try to reload, it doesn't warn, yes. But this query will be the last in history

Thats correct behaviour.

> 2\. if active window isn't `QueryEditor`. I'm not sure, that we should warn in such cases

We should warn that time, when unsaved content in editor is going to be lost. (reseted or replaced) 

The usecase: I'd written something in the editor (I had not run it), and after that I moved to the cluster's root and to another database. Content in editor is lost, but I was't warned about it :(

artemmufazalov:
> We should warn that time, when unsaved content in editor is going to be lost. (reseted or replaced)
> 
> The usecase: I'd written something in the editor (I had not run it), and after that I moved to the cluster's root and to another database. Content in editor is lost, but I was't warned about it :(

It seems like a big issue with several scenarios:
1. Top Queries, History and Saved Tabs
2. Schema actions (create, select, upsert and etc.)
3. Navigation from Tenant page
4. Browser actions (page reload)

I couldn't figure out fast and easy solution, so I'd prefer to tackle this issue separately. There is a ticket for it already, I'll send it to you in PM

Raubzeug (approved):

PR #439, opened 2023-06-26
fix(Tenants): fix tenant link

artemmufazalov:


src/containers/Tenants/Tenants.js
```diff
@@ -154,7 +147,6 @@ class Tenants extends React.Component {
                                     name: value,
                                     backend,
                                     [TenantTabsGroups.info]: initialTenantInfoTab,
-                                    [TenantTabsGroups.general]: initialTenantGeneralTab,
```
artemmufazalov, line 157: If undefined, it's taken from settings in tenant page. There is no need to define it explicitly

Raubzeug (approved):

Raubzeug (approved):

PR #448, opened 2023-06-29
fix(Storage): remove visible entities filter

artemmufazalov:
After #441

### DIFF
--- a/src/store/reducers/storage/storage.js
+++ b/src/store/reducers/storage/storage.js
@@ -335,19 +335,6 @@ export const getFlatListStorage = createSelector(
     },
 );
 
-export const getVisibleEntitiesList = createSelector(
-    [getVisibleEntities, getFlatListStorage],
-    (visibleGroups, storageList) => {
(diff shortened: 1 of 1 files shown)


src/store/reducers/storage/storage.js
```diff
@@ -335,19 +335,6 @@ export const getFlatListStorage = createSelector(
...
-        if (visibleGroups === VISIBLE_ENTITIES.all) {
-            return storageList;
-        } else if (visibleGroups === VISIBLE_ENTITIES.missing) {
-            return _.filter(storageList, (g) => g.Missing > 0);
-        } else {
-            return _.filter(storageList, (g) => g.UsedSpaceFlag > 100);
-        }
-    },
-);
-
```
artemmufazalov: It should be filtered on backend. Additionally, currently it's incorrect, that causes an issue, where no out of space nodes are displayed

Raubzeug (approved):

PR #461, opened 2023-07-05
fix: show 5 digits size in table info

artemmufazalov:
Screen Shot 2023-07-05 at 17 36 50


src/utils/bytesParsers/formatBytes.ts
```diff
@@ -0,0 +1,131 @@
...
+    value,
+    size = 'mb',
+    precision = 0,
+    withLabel = true,
+    isSpeed = false,
+}: FormatToSizeArgs) => {
+    let result = (Number(value) / sizes[size].value).toFixed(precision);
+    result = formatNumber(result);
+
+    if (withLabel) {
```
Raubzeug, line 50: Lets move this case and case with `isSpeed` to another functions. Now it's pretty unclear what for this function is responsible.

src/utils/bytesParsers/__test__/formatBytes.test.ts
```diff
@@ -0,0 +1,31 @@
+import {formatBytes} from '../formatBytes';
+
+describe('formatBytes', () => {
```
Raubzeug, line 3: Lets add test-cases with invalid data.

Raubzeug (changes requested):

src/utils/bytesParsers/formatBytes.ts
artemmufazalov: Do you suggest separating conditions? Or there shouldn't be logic with adding label in such function?

Raubzeug: I see it as two additional functions: one will give simple label, one will give `/s` label.

```
function withSpeed(val) {
return withLabel(val)
}
function withLabel(val) {
return formatBytes(val)
}
```

src/utils/bytesParsers/formatBytes.ts
```diff
@@ -105,27 +69,64 @@ const formatToSignificantDigits = ({
...
+};
 
 /**
  * @param significantDigits - number of digits above 3
  */
-export const formatBytes = ({value, size, ...params}: FormatBytesArgs) => {
+export const formatBytes = ({
+    value,
+    size,
+    withSpeedLabel = false,
```
Raubzeug, line 108: I meant a little bit different. I suggested not to have this props in this function: instead of passing this props let's call different functions - `formatSpeedBytes` and `formatSizeBytes`. Inside this two functions `formatBytes` should be called.

artemmufazalov: In this case I wanted to create universal function to format bytes, so it was made intentionally. And for me additional `withSpeedLabel` and `withSizeLabel` params don't make it purpose unclear

Raubzeug: It's up to you! As for me - the less `if`-s - the better :)

Raubzeug (approved):

PR #471, opened 2023-07-12
chore: update types

artemmufazalov:


Raubzeug:
We discussed, that its better not to use enums. It up to you, I just wanted to remind this.

Raubzeug (approved):

artemmufazalov:
These types are taken from backend proto files. There are enums there, when I make TS types I leave enums as they are, because in my view it makes easier to view, compare and update them. Regarding self-written types, I took your recommendation and use JS objects

PR #476, opened 2023-07-17
fix(Nodes): support v2 compute

artemmufazalov:


src/types/api/nodes.ts
```diff
@@ -103,12 +103,13 @@ enum EConfigState {
...
-    DC,
-    Version,
-    Uptime,
-    Memory,
-    CPU,
-    LoadAverage,
+    NodeId = 'NodeId',
+    Host = 'Host',
+    DC = 'DC',
+    Rack = 'Rack',
```
artemmufazalov, line 109: Added `Rack`. Now this `ESort` is the same as `ESort` for `/nodes`, but I decided to have two different enums since endpoints are different and they are handled differently

Raubzeug (approved):

PR #480, opened 2023-07-21
fix(Query): process null issues error

artemmufazalov:


src/containers/Tenant/Query/Issues/Issues.tsx
```diff
@@ -123,18 +120,24 @@ function Issue({issue, level = 0}: {issue: IssueType; expanded?: boolean; level?
...
-                {issue.code ? <span className={blockIssue('code')}>Code: {issue.code}</span> : null}
+                {issue.issue_code ? (
+                    <span className={blockIssue('code')}>Code: {issue.issue_code}</span>
+                ) : null}
             </div>
             {hasIssues && isExpand && (
                 <div className={blockIssue('issues')}>
-                    <IssueList issues={issue.issues!} level={level + 1} expanded={isExpand} />
+                    <IssueList
+                        issues={issue.issues as IssueMessage[]}
```
Raubzeug, line 130: Maybe it is possible to check if `issue.issues` are valid instead of type casting?

Raubzeug (approved):

PR #482, opened 2023-07-24
fix(ObjectSummary): fix issue on object change with active schema tab

artemmufazalov:


src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -143,13 +143,10 @@ export function ObjectSummary({
...
 
-        if (type && !isTable && !TENANT_INFO_TABS.find((el) => el.id === infoTab)) {
-            history.push({
-                pathname: location.pathname,
-                search: qs.stringify({...queryParams, info: TenantInfoTabsIds.overview}),
-            });
+        if (type && !isTable && !TENANT_INFO_TABS.find((el) => el.id === summaryTab)) {
+            dispatch(setSummaryTab(TENANT_SUMMARY_TABS_IDS.overview));
         }
-    }, [type, history, infoTab, location, queryParams]);
```
artemmufazalov, line 152: There is an issue, when Schema tab is selected and object without Schema tab is clicked, nothing happens. That happens because data in store is updated before query params. So `type` and `infoTab` are updated, `useEffect` is called, new query is pushed to `history`, but with old schema path, so object in navigation tree doesn't change.

It could be fixed with explicit schema tab param, but this won't fix the root issue of values that dependant from each other but don't update synchronously.

So to fix it I decided to put everything in store, so `summaryTab` is updated together with other store values, like `type` or `currentSchemaPath`

Raubzeug (approved):

PR #483, opened 2023-07-24
fix: fix copy schema action

artemmufazalov:


src/containers/Tenant/utils/schemaActions.ts
```diff
@@ -51,22 +52,20 @@ const bindActions = (
         selectQuery: inputQuery(selectQueryTemplate),
         upsertQuery: inputQuery(upsertQueryTemplate),
         copyPath: () => {
-            navigator.clipboard
```
artemmufazalov, line 54: `navigator.clipboard` was not always available for embedded version, so sometimes copy didn't work. Replaced it with `copy` from `copy-to-clipboard` lib, that is used in `react-copy-to-clipboard`, that is used in `uikit`, so similar behaviour is expected

src/containers/Tenant/utils/schemaActions.ts
```diff
@@ -1,15 +1,16 @@
 import {Dispatch} from 'react';
+import copy from 'copy-to-clipboard';
```
Raubzeug, line 2: Why did you decide not to use `CopyToClipboard` from `uikit`?

artemmufazalov: Because there is a component in `uikit`, but in this case I needed only `copy` action to add it to an option in schema tree

Raubzeug (approved):

PR #484, opened 2023-07-24
chore: fix pre-commit prettier linting and add json linting

artemmufazalov:
Return changes reverted in #197


.eslintrc
```diff
@@ -1,12 +1,9 @@
 {
-    "extends": ["@gravity-ui/eslint-config", "prettier"],
```
artemmufazalov, line 2: There was an error, prettier was not applied

Raubzeug (approved):

PR #490, opened 2023-07-31
fix: fix tablet size

krosy1337:
image

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -1,5 +1,7 @@
 import DataTable from '@gravity-ui/react-data-table';
 
+import {EType} from '../types/api/tablet';
+
 const SECOND = 1000;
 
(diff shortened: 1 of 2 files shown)


src/utils/constants.ts
```diff
@@ -51,6 +51,7 @@ export const TABLET_SYMBOLS = {
     BlockStoreVolume: 'BV',
     Console: 'CN',
     TenantSlotBroker: 'TB',
+    BlockStoreDiskRegistry: 'BDR',
```
artemmufazalov, line 54: Could you please use `EType` enum values from `/types/api/tablet` instead of string keys here? I think it will be more error safe:
```
[EType.OldTxProxy]: "P",
```

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #491, opened 2023-07-31
fix(SchemaTree): set required query mode for tree actions

artemmufazalov:
https://github.com/ydb-platform/ydb-embedded-ui/pull/485#issuecomment-1653606646


src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
```diff
@@ -21,6 +24,9 @@ export function SchemaTree(props: SchemaTreeProps) {
 
     const dispatch = useDispatch();
 
+    const [_, setQueryMode] = useSetting<QueryMode>(QUERY_INITIAL_MODE_KEY);
+    const [enableAdditionalQueryModes] = useSetting<boolean>(ENABLE_ADDITIONAL_QUERY_MODES);
```
Raubzeug, line 28: what do you think about combining it into one hook, which will determine available modes and show an error toast if unavailable mode is going to be set.

src/containers/Tenant/utils/schemaActions.ts
```diff
@@ -34,23 +35,44 @@ const upsertQueryTemplate = (path: string) => {
...
 
+interface ActionsOptions {
+    enableAdditionalQueryModes?: boolean;
+    setQueryMode?: (mode: QueryMode) => void;
+}
+
 const bindActions = (
     path: string,
     dispatch: Dispatch<any>,
     setActivePath: (path: string) => void,
```
Raubzeug, line 46: it seems that `setActivePath` should be moved to `options`.

src/containers/Tenant/utils/schemaActions.ts
```diff
@@ -34,23 +35,44 @@ const upsertQueryTemplate = (path: string) => {
...
+    options?: ActionsOptions,
```
Raubzeug, line 47: I'd better rename this prop to `additionalEffects` maybe...

src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
artemmufazalov: I'm not sure that it's currently needed. 
1) Is will be used only here, because in `QueryEditor` we will need both values to manage controls
2) It's a temporary solution while there is no ydb versions where QueryService is supported (it was supposed to be in 23-2, but 2 months ago we forgot to merge PR)

src/containers/Tenant/utils/schemaActions.ts
artemmufazalov: I don't like `additionalEffects` name. I'll call it just `additionalParams`

src/containers/Tenant/Schema/SchemaTree/SchemaTree.tsx
Raubzeug: I suggested this, because I don't like passing `enableAdditionalQueryModes` to `bindActions`. I think that the decision to change mode or not to change should be taken here, not in `bindActions`

artemmufazalov: Now I've got your point. I'll change it

Raubzeug (approved):

PR #500, opened 2023-08-07
fix(Header): add icons for nodes and tablets

krosy1337:
image
image
image

### DIFF
--- a/src/components/TabletIcon/TabletIcon.scss
+++ b/src/components/TabletIcon/TabletIcon.scss
@@ -2,19 +2,16 @@
     display: flex;
     justify-content: center;
 
-    width: 16px;
+    width: 23px;
     height: 16px;
 
-    font-size: 6px;
-    line-height: 13px;
-    font-weight: 600;
+    font-size: 10px;
 
-    border: 2px solid var(--yc-color-text-primary);
+    border-width: 1px;
+    border-style: solid;
     border-radius: 4px;
 
-    &__square {
-        position: absolute;
-        top: 0;
-        left: 0;
+    &__type {
+        line-height: 14px;
     }
 }
--- a/src/components/TabletIcon/TabletIcon.scss
+++ b/src/components/TabletIcon/TabletIcon.scss
@@ -8,8 +8,7 @@
     font-size: 10px;
     text-transform: uppercase;
 
-    border-width: 1px;
-    border-style: solid;
+    border: 1px solid;
     border-radius: 4px;
 
     &__type {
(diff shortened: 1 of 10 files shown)


src/containers/Tablet/Tablet.tsx
```diff
@@ -48,8 +48,9 @@ export const Tablet = () => {
         error,
     } = useTypedSelector((state) => state.tablet);
 
-    const {nodeId: queryNodeId, tenantName: queryTenantName} = parseQuery(location);
+    const {TabletId, Overall, Leader, Type} = tablet;
 
+    const {nodeId: queryNodeId, tenantName: queryTenantName} = parseQuery(location);
```
artemmufazalov, line 53: Maybe `Type` additionally could be passed via search params so it would be rendered immediately? Now we have placeholder `T` until data is fully fetched.

`Type` from tablet data will be used only if there is no `Type` in search. To Tablet page we can get only from `Tablet` component, so there all params are set

artemmufazalov (changes requested):

src/containers/Tablet/Tablet.tsx
```diff
@@ -48,10 +49,14 @@ export const Tablet = () => {
         error,
     } = useTypedSelector((state) => state.tablet);
 
-    const {nodeId: queryNodeId, tenantName: queryTenantName} = parseQuery(location);
-
+    const {
+        nodeId: queryNodeId,
+        tenantName: queryTenantName,
+        type: queryType,
```
artemmufazalov, line 55: I'd call it `queryTabletType` to ensure more clear naming

src/containers/Tablet/Tablet.tsx
```diff
@@ -48,10 +49,14 @@ export const Tablet = () => {
...
+    } = parseQuery(location);
     const nodeId = tablet.NodeId?.toString() || queryNodeId?.toString();
     const tenantName = tenantPath || queryTenantName?.toString();
+    const type = (tablet.Type?.toString() as EType) || (queryType?.toString() as EType);
```
artemmufazalov, line 59: You need parse to string only type from query string, `tablet.Type` already has correct type.
And the type for `queryType` would be `EType | undefined`

src/components/TabletIcon/TabletIcon.scss
```diff
@@ -0,0 +1,17 @@
+.tablet-icon {
```
artemmufazalov, line 1: There is some issue with align (probably due to `line-height` property). It seems the labels don't align by the center of a tablet

Screen Shot 2023-08-09 at 13 35 01

artemmufazalov (changes requested):

PR #503, opened 2023-08-07
feat(Nodes): filter and sort on backend

artemmufazalov:
Screen Shot 2023-08-07 at 17 51 48

### DIFF
--- a/src/containers/Nodes/Nodes.tsx
+++ b/src/containers/Nodes/Nodes.tsx
@@ -99,12 +99,14 @@ export const Nodes = ({path, type, additionalNodesInfo = {}}: NodesProps) => {
                 dispatch(setDataWasNotLoaded());
             }
 
+            const params = requestParams || {};
+
             // For not DB entities we always use /compute endpoint instead of /nodes
             // since /nodes can return data only for tenants
             if (path && (!useNodesEndpoint || !isDatabaseEntityType(type))) {
-                dispatch(getComputeNodes({path, ...requestParams}));
+                dispatch(getComputeNodes({path, ...params}));
             } else {
-                dispatch(getNodes({tenant: path, ...requestParams}));
+                dispatch(getNodes({tenant: path, ...params}));
             }
         },
         [dispatch, path, type, useNodesEndpoint, requestParams],
--- a/src/containers/Nodes/Nodes.tsx
+++ b/src/containers/Nodes/Nodes.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useMemo, useState} from 'react';
+import {useCallback, useEffect, useMemo} from 'react';
 import cn from 'bem-cn-lite';
 import {useDispatch} from 'react-redux';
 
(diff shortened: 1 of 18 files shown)


src/containers/Nodes/Nodes.tsx
```diff
@@ -58,27 +72,88 @@ export const Nodes = ({path, type, additionalNodesInfo = {}}: NodesProps) => {
...
+            } else {
+                dispatch(getNodes({tenant: path, ...requestParams}));
+            }
+        },
+        [dispatch, path, type, useNodesEndpoint, requestParams],
+    );
 
     useAutofetcher(fetchNodes, [fetchNodes], isClusterNodes ? true : autorefresh);
 
+    const sort: SortOrder | undefined = useMemo(() => {
```
Raubzeug, line 141: If we are going to use different sort types in other places, it may be useful to move calculating sort params (and maybe even `handleSort`) into a hook or a helper.

Raubzeug (changes requested):

artemmufazalov:
> When change sort order or filters with backend, Skeleton is obtrusively blinking.

It blinks because data is loaded too fast. If it is some bigger cluster, skeletons would be just fine.

Raubzeug:
> It blinks because data is loaded too fast. If it is some bigger cluster, skeletons would be just fine.

Just thinking... What do you think about not showing skeleton but highlight that data is deprecated (maybe font color -> secondary)?

src/utils/hooks/useNodesRequestParams.ts
```diff
@@ -0,0 +1,49 @@
...
+    sortOrder,
+    sortValue,
+}: NodesRawRequestParams) => {
+    const [useBackendParamsForTables] = useSetting<boolean>(USE_BACKEND_PARAMS_FOR_TABLES_KEY);
+
+    const [requestParams, setRequestParams] = useState<NodesGeneralRequestParams>({});
+
+    // If backend params are enabled, update params value to use them in fetch request
+    // Otherwise no params will be updated, no hooks that depend on requestParams will be triggered
+    useEffect(() => {
```
Raubzeug, line 30: It seems that you don't need `useEffect` and `useState` here.
If `useSetting` rerenders -> `useNodesRequestParams` rerenders and recalculate `request` params in a base flow.

artemmufazalov: It will refetch data on every component rerender (that could be caused not only by request params change).
I replaced two hooks with `useMemo` (additionally also fixed the bug, that params weren't reset on setting toggle)

Raubzeug:
> It looks good for fast loads, but weird when loading takes some time.
> 
> I think this should be discussed with the team and maybe our designer, because such changes should relate to all tables.

Ok. I'm talking about it because in users perception sorting is not refetching (even if it is, but user is not aware of realisation of an app). And thats why it's weird to see a skeleton here.

Raubzeug (approved):

PR #506, opened 2023-08-09
feat(Storage): v2 api and backend filters

artemmufazalov:
To turn in backend filters you need to toggle the setting. Currently it affects only filters, sorting will be added further. 
 
Screen Shot 2023-08-14 at 14 23 15


src/utils/hooks/useStorageRequestParams.ts
```diff
@@ -0,0 +1,21 @@
+import {useMemo} from 'react';
+
+import {EVersion} from '../../types/api/storage';
+import {USE_BACKEND_PARAMS_FOR_TABLES_KEY} from '../constants';
+import {useSetting} from './useSetting';
+
+export const useStorageRequestParams = ({filter}: {filter?: string}) => {
```
artemmufazalov, line 7: Sorting params and proper params type will be added further similar to nodes

src/utils/hooks/useStorageRequestParams.ts
```diff
@@ -0,0 +1,21 @@
...
+    const [useBackendParamsForTables] = useSetting<boolean>(USE_BACKEND_PARAMS_FOR_TABLES_KEY);
+
+    // If backend params are enabled, update params value to use them in fetch request
+    // Otherwise no params will be updated, no hooks that depend on requestParams will be triggered
+    return useMemo(() => {
+        if (useBackendParamsForTables) {
+            return {
+                version: EVersion.v2,
```
artemmufazalov, line 15: Currently `v2` storage has some issues, so it will be used only when params enabled, as only `v2` works with sorting and filters

Raubzeug (approved):

PR #509, opened 2023-08-15
feat(Diagnostics): update Healthcheck design

krosy1337:
image
image


src/components/DiagnosticCard/DiagnosticCard.tsx
```diff
@@ -0,0 +1,17 @@
...
+import './DiagnosticCard.scss';
+
+const b = cn('diagnostic-card');
+
+interface DiagnosticCardProps {
+    children?: ReactNode;
+    className?: string;
+}
+
+export function DiagnosticCard({children, className}: DiagnosticCardProps) {
```
artemmufazalov, line 15: Why do you need this component? It seems you can just use a card and set it styles in `TenantOverview`. Or do you plan to use it for cards with metrics?

src/components/EntityStatus/EntityStatus.scss
```diff
@@ -89,6 +106,13 @@
             width: 18px;
             height: 18px;
         }
+
+        &_size_l {
+            aspect-ratio: 1;
```
artemmufazalov, line 111: It seems this isn't needed

src/components/EntityStatus/EntityStatus.scss
```diff
@@ -47,6 +47,23 @@
...
+        }
+        &_state_yellow {
+            color: var(--yc-color-infographics-warning-heavy);
+        }
+        &_state_orange {
+            color: var(--yc-color-base-warning-orange);
+        }
+        &_state_red {
+            color: var(--yc-color-infographics-danger-heavy);
+        }
```
artemmufazalov, line 66: There is already classes with colors for status icons (`&__status-color { &_state_green {} }`. Can you somehow reuse them?

src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
```diff
@@ -41,7 +41,10 @@ export const Details = (props: DetailsProps) => {
     const renderHealthcheckIssues = () => {
         return (
             <div className={b('issues-wrapper')}>
-                <IssueTree issueTree={issueTree} />
+                {issueTrees.map((issueTree) => {
+                    if (issueTree) return <IssueTree key={issueTree.id} issueTree={issueTree} />;
+                    return undefined;
```
artemmufazalov, line 46: It seems in this case `issueTree` is always defined and this check is redundant

src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
```diff
@@ -13,15 +13,15 @@ import i18n from '../i18n';
...
     loading?: boolean;
     onUpdate: VoidFunction;
 }
 
 export const Details = (props: DetailsProps) => {
-    const {loading, onUpdate, issueTree} = props;
+    const {loading, onUpdate, issueTrees} = props;
 
-    if (!issueTree) {
+    if (!issueTrees) {
```
artemmufazalov, line 24: Is seems that `issueTrees` are always and array, but this array could be empty (`[]`). So it better to check `if (!issueTrees || !issuesTrees.length) {`

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.scss
```diff
@@ -67,6 +69,17 @@
         line-height: 24px;
     }
 
+    &__issues {
```
artemmufazalov, line 72: Can you call this class somehow differently? They are not the issues itself, but their count

artemmufazalov (changes requested):

src/components/DiagnosticCard/DiagnosticCard.tsx
krosy1337: Yes, I'm going to use this component for cards with metrics.

src/containers/Tenant/Diagnostics/Healthcheck/Preview/Preview.tsx
```diff
@@ -39,33 +48,44 @@ export const Preview = (props: PreviewProps) => {
...
+                                    label={count.toString()}
+                                    size="l"
+                                />
+                            ))}
+                        </div>
+                        <Link
+                            onClick={() => onShowMore?.(issuesTreesRoots.map((issue) => issue.id))}
+                        >
+                            {i18n('label.show-details')}
+                        </Link>
```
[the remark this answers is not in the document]
artemmufazalov: Using `Button` implies some additional logic, so decided to left `Link` as it was before

src/store/reducers/healthcheckInfo.ts
```diff
@@ -121,14 +140,20 @@ export const selectIssuesTrees: Selector<IHealthcheckInfoRootStateSlice, IIssues
...
     IHealthcheckInfoRootStateSlice,
-    IIssuesTree | undefined,
-    [string | undefined]
-> = createSelector([selectIssuesTrees, (_, id: string | undefined) => id], (issuesTrees = [], id) =>
-    issuesTrees.find((issuesTree) => issuesTree.id === id),
+    IIssuesTree[],
+    [string[] | undefined]
+> = createSelector(
+    [selectIssuesTrees, (_, ids: string[] | undefined) => ids],
+    (issuesTrees = [], ids) => issuesTrees.filter((issuesTree) => ids?.includes(issuesTree.id)),
```
artemmufazalov, line 149: It's no longer used

src/store/reducers/healthcheckInfo.ts
```diff
@@ -110,6 +109,26 @@ const getInvertedConsequencesTree = ({
...
+    const sortedStatusFlags: StatusFlag[] = _flow([
+        _sortBy((status: StatusFlag) => mapStatusToPriority[status]),
+    ])(Object.keys(issuesMap));
+
+    return sortedStatusFlags.reduce((obj, key) => {
+        obj[key] = issuesMap[key];
+        return obj;
+    }, {} as Record<StatusFlag, number>);
+};
+
```
artemmufazalov, line 131: Do you really need to prepare statistics as `Record<StatusFlag, number>`? It seems you don't use it anywhere else and transform before render. So you can just return `[StatusFlag, number][]` and use it as it is in component
```
return Object.entries(issuesMap).sort(sortFunc)
```

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/Healthcheck/Details/Details.tsx
```diff
@@ -38,18 +36,28 @@ export const Details = (props: DetailsProps) => {
...
         );
     };
 
     return (
         <div className={b('details')}>
             {renderHealthcheckHeader()}
-            {renderHealthcheckIssues()}
+            {renderContent()}
         </div>
     );
```
artemmufazalov, line 62: It will be better to wrap error and no data message into `issues-wrapper` css class to remain constant height with different content. 

Probably you could do something like this and put styles from `issues-wrapper` to `details-content-wrapper`
```
 return (
        <div className={b('details')}>
            {renderHealthcheckHeader()}
            <div className={b('details-content-wrapper')}>{renderContent()}</div>
        </div>
    );
```
Styles I'm writing about
```
&__issues-wrapper {
        overflow-x: hidden;
        overflow-y: auto;

        height: 70vh;
        max-height: 70vh;
    }
```

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #519, opened 2023-08-23
refactor: migrate TenantOverview to ts

krosy1337:


src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.tsx
```diff
@@ -0,0 +1,156 @@
...
+        [dispatch, tenantName],
+    );
+
+    useAutofetcher(
+        (isBackground) => {
+            fetchTenant(isBackground);
+        },
+        [fetchTenant],
+        autorefresh,
+    );
```
artemmufazalov, line 48: You can just `useAutofetcher(fetchTenant, [fetchTenant], autorefresh]`

src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.tsx
```diff
@@ -0,0 +1,156 @@
...
+    const memoryRaw = MemoryUsed ?? Metrics.Memory;
+
+    const memory = (memoryRaw && bytesToGB(memoryRaw)) || 'no data';
+    const storage = (Metrics.Storage && bytesToGB(Metrics.Storage)) || 'no data';
+    const storageGroups = StorageGroups ?? 'no data';
+    const blobStorage = (StorageAllocatedSize && bytesToGB(StorageAllocatedSize)) || 'no data';
+    const storageEfficiency =
+        Metrics.Storage && StorageAllocatedSize
+            ? `${((parseInt(Metrics.Storage) * 100) / parseInt(StorageAllocatedSize)).toFixed(2)}%`
+            : 'no data';
```
artemmufazalov, line 73: It's better to create a variable for placeholder `no data`. And to use `i18n` for it

artemmufazalov, line 73: Note for future refactoring: there are helpers in 
https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/store/reducers/tenants/utils.ts

It's better to use them in `tenant.ts` reducer as well and use prepared data here

There only difference is that there are `Tablet storage` and `Blob storage` here and there we have single storage that is one of these values

src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.tsx
```diff
@@ -0,0 +1,156 @@
...
+        {label: 'Storage groups', value: storageGroups},
+        {label: 'Blob storage', value: blobStorage},
+        {label: 'Storage efficiency', value: storageEfficiency},
+    ];
+
+    const tabletsInfo = StateStats.map((info) => {
+        if (info.VolatileState)
+            return {label: TABLET_STATES[info.VolatileState], value: info.Count};
+        return {};
+    });
```
artemmufazalov, line 93: It's better to filter values and use type guard inside filter, than to return an empty object. Also there won't be any need in making `InfoViewer` `label` optional

`StateStats.filter().map()`

src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    const renderName = (tenant?: TTenant) => {
+        const {Name = TENANT_DEFAULT_TITLE, State} = tenant || {};
```
artemmufazalov, line 96: You can use data that destructured on lines 50-61 (but introduce `TENANT_DEFAULT_TITLE` only here)
```
    const renderName = () => {
        return (
            <div className={b('tenant-name-wrapper')}>
                <EntityStatus
                    status={State}
                    name={Name || TENANT_DEFAULT_TITLE}
                    withLeftTrim
                    hasClipboardButton={tenant}
                    clipboardButtonAlwaysVisible
                />
            </div>
        );
    };
```

src/components/InfoViewer/InfoViewer.tsx
```diff
@@ -4,8 +4,8 @@ import cn from 'bem-cn-lite';
 import './InfoViewer.scss';
 
 export interface InfoViewerItem {
-    label: ReactNode;
-    value: ReactNode;
+    label?: ReactNode;
```
artemmufazalov, line 7: `label` shouldn't be optional

artemmufazalov (changes requested):

src/store/reducers/tenants/utils.ts
```diff
@@ -2,7 +2,7 @@ import type {TTenant} from '../../../types/api/tenant';
 import {isNumeric} from '../../../utils/utils';
 
 const getControlPlaneValue = (tenant: TTenant) => {
-    const parts = tenant.Name.split('/');
+    const parts = tenant.Name ? tenant.Name.split('/') : [];
```
artemmufazalov, line 5: It can be simplified 
```
    const parts = tenant.Name?.split('/');
    const defaultValue = parts?.length ? parts[parts.length - 1] : '—';
```

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #527, opened 2023-08-29
feat: update TenantOverview design

krosy1337:
image


Raubzeug (changes requested):

artemmufazalov (changes requested):

src/components/DiagnosticCard/DiagnosticCard.tsx
```diff
@@ -5,17 +5,17 @@ import {Card} from '@gravity-ui/uikit';
...
     className?: string;
-    selected?: boolean;
+    active?: boolean;
 }
 
-export function DiagnosticCard({children, className, selected}: DiagnosticCardProps) {
+export function DiagnosticCard({children, className, active}: DiagnosticCardProps) {
     return (
-        <Card view="clear" className={b({selected}, className)}>
+        <Card view="clear" className={b({active}, className)}>
```
krosy1337, line 18: There are problems with using type="action" because it is difficult to overwrite styles from Gravity UI

src/containers/Tenant/Diagnostics/Healthcheck/Healthcheck.tsx
```diff
@@ -77,6 +78,7 @@ export const Healthcheck = (props: HealthcheckProps) => {
                 onShowMore={showMoreHandler}
                 onUpdate={fetchHealthcheck}
                 error={error}
+                selected={selected}
```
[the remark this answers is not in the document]
krosy1337: It doesn't seem worth adding context for a single prop

src/containers/Tenant/Diagnostics/TenantOverview/MetricsCards/MetricsCards.tsx
```diff
@@ -0,0 +1,66 @@
...
+        cpu: cpuUsed,
+        storage: storageUsed,
+        memory: memoryUsed,
+    });
+
+    return (
+        <div className={b()}>
+            <MetricCard label="CPU" progress={cpuUsage} status={cpuStatus} resourcesUsed={cpu} />
+            <MetricCard
+                label="Storage"
```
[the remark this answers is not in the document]
krosy1337: We decided not to add a translation for cards titles

src/components/DiagnosticCard/DiagnosticCard.tsx
```diff
@@ -10,8 +10,13 @@ const b = cn('diagnostic-card');
...
     children?: ReactNode;
     className?: string;
+    selected?: boolean;
 }
 
-export function DiagnosticCard({children, className}: DiagnosticCardProps) {
-    return <Card className={b(null, className)}>{children}</Card>;
+export function DiagnosticCard({children, className, selected}: DiagnosticCardProps) {
+    return (
+        <Card view="clear" className={b({selected}, className)}>
```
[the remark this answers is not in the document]
krosy1337: There are problems with using type="action" because it is difficult to overwrite styles for background and shadows from Gravity UI

src/utils/index.js
```diff
@@ -109,3 +109,14 @@ export const renderExplainNode = (node) => {
     const parts = node.name.split('|');
     return parts.length > 1 ? parts[1] : node.name;
 };
+
+export const normalizeProgress = (progress) => {
```
Raubzeug, line 113: No need to move this function to common utils, because it is used only in one file.

src/components/DiagnosticCard/DiagnosticCard.tsx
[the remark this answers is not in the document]
Raubzeug: If you need such a big overwriting why did you decide to use `Card` from uikit? What benefits do you get from using it?

src/components/DiagnosticCard/DiagnosticCard.scss
```diff
@@ -1,5 +1,25 @@
-.diagnostic-card {
-    min-width: 200px;
-    max-width: 350px;
+.ydb-diagnostic-card {
+    flex-shrink: 0;
+
+    width: 206px;
```
Raubzeug, line 4: it seems that height should be fixed too.

src/containers/Tenant/Diagnostics/Healthcheck/Preview/Preview.tsx
```diff
@@ -75,8 +78,8 @@ export const Preview = (props: PreviewProps) => {
     };
 
     return (
-        <DiagnosticCard className={b('preview')}>
-            {renderStatus()}
+        <DiagnosticCard className={b('preview')} active={active}>
```
Raubzeug, line 81: If new view is not activated, Healthcheck should not be shown in the new way.

Raubzeug, line 81: Did you check layout if status is not ok?

Raubzeug (changes requested):

src/store/reducers/tenants/types.ts
```diff
@@ -30,3 +32,5 @@ export type TenantsAction =
 export interface TenantsStateSlice {
     tenants: TenantsState;
 }
+
+export type MetricStatus = ValueOf<typeof METRIC_STATUS>;
```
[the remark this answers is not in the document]
artemmufazalov: We get types from values throughout our project. That's why type `ValueOf` was created. The reasons: we pass values as arguments, not the keys

src/containers/Tenant/Diagnostics/Healthcheck/Preview/Preview.tsx
artemmufazalov: New Healthcheck looks alright with the old design, I don't understand the problem here.

Additionally, we changed not only preview, but also modal content. For the new modal old preview doesn't suits anymore (we show all issues in preview at once, not only selected issue)

src/components/DiagnosticCard/DiagnosticCard.scss
artemmufazalov: It isn't fixed because of Healthcheck preview. In some rare cases there could be a lot of issues of different types in database, in this case HC Preview will be bigger and Card should grow in terms of height. 

We won't apply height that will cover all the possible scenarios, as it doesn't look as good as the current design. It's a very rare case, but we want content to be shown normally in all the cases.

src/containers/Tenant/Diagnostics/Healthcheck/Preview/Preview.tsx
Raubzeug: As for me, if the new design is under construction, the old one should be untouched.
But it's up to the team of course!

Raubzeug (approved):

artemmufazalov (approved):

artemmufazalov (approved):

PR #528, opened 2023-08-30
fix: fix Healthcheck blinking

krosy1337:


src/services/api.ts
```diff
@@ -350,10 +350,14 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
...
-        return this.get<HealthCheckAPIResponse>(this.getPath('/viewer/json/healthcheck'), {
-            tenant: database,
-        });
+    getHealthcheckInfo(database: string, {concurrentId}: AxiosOptions = {}) {
+        return this.get<HealthCheckAPIResponse>(
+            this.getPath('/viewer/json/healthcheck'),
+            {
+                tenant: database,
+            },
+            {concurrentId: concurrentId || `getHealthcheckInfo|${database}`},
```
artemmufazalov, line 359: Default name isn't correct here as it doesn't fix bug with blinking. In this case it's better not to pass default name at all and left `{tenant: database}, {concurrentId}`

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #531, opened 2023-09-13
feat: move healthcheck to tabs

krosy1337:


artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/TenantOverview/MetricsCards/MetricsCards.scss
artemmufazalov, line 1: There is a problem with tabs in Safari. Some of our users use it, so we should consider it as well.

1. Metrics are invisible
2. When you select and unselect card there is some flash 

Screen Shot 2023-09-15 at 17 04 38

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/TenantOverview/Healthcheck/Healthcheck.scss
```diff
@@ -0,0 +1,110 @@
+@use '../../DetailedOverview/DetailedOverview.scss' as detailedOverview;
```
artemmufazalov, line 1: It's better to move variables from `DetailedOverview.scss` to `src/styles/constants.scss`. It makes their usage more clear. They could be named like `--diagnostics-section-title-margin`

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #539, opened 2023-09-22
fix(Authentication): enable page redirect

artemmufazalov:
Fix the bug with `Authentication`. When user logins manually, after login click page doesn't redirect to the initial url

### DIFF
--- a/src/containers/AsideNavigation/AsideNavigation.tsx
+++ b/src/containers/AsideNavigation/AsideNavigation.tsx
@@ -50,7 +50,9 @@ function YbdInternalUser({ydbUser, logout}: YbdInternalUserProps) {
     const history = useHistory();
 
     const handleLoginClick = () => {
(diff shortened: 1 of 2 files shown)


src/containers/AsideNavigation/AsideNavigation.tsx
```diff
@@ -50,7 +50,9 @@ function YbdInternalUser({ydbUser, logout}: YbdInternalUserProps) {
     const history = useHistory();
 
     const handleLoginClick = () => {
-        history.push(createHref(routes.auth, undefined, {returnUrl: encodeURI(location.href)}));
+        history.push(
+            createHref(routes.auth, undefined, {returnUrl: encodeURIComponent(location.href)}),
```
artemmufazalov, line 54: Use `encodeURIComponent` instead `encodeURI`, because the latest doesn't encode hostname

Raubzeug (approved):

PR #541, opened 2023-09-26
feat(TenantOverview): add storage tab to tenant diagnostics

krosy1337:


src/utils/dataFormatters/dataFormatters.ts
```diff
@@ -55,11 +60,26 @@ export const formatMsToUptime = (ms?: number) => {
     return ms && formatUptime(ms / 1000);
 };
 
-export const formatStorageValues = (value?: number, total?: number) => {
-    return [
-        value ? String(Math.floor(value / TERABYTE)) : undefined,
-        total ? `${Math.floor(total / TERABYTE)} TB` : undefined,
-    ];
+export const formatStorageValues = (value?: number, total?: number, size?: BytesSizes) => {
```
artemmufazalov, line 63: I think this function is too multi purpose. I don't understand what exactly it should do and how it should process data. I think it should be split apart

artemmufazalov (changes requested):

artemmufazalov (changes requested):

artemmufazalov:
Good job! Almost everything is fixed)

artemmufazalov (approved):

PR #542, opened 2023-10-05
fix: fix cpu usage calculation

krosy1337:


src/store/reducers/tenants/utils.ts
```diff
@@ -22,36 +22,28 @@ const getTenantBackend = (tenant: TTenant) => {
...
-
-    const memory = isNumeric(rawMemory) ? Number(rawMemory) : undefined;
-
-    // Blob storage - actual database size
+    const cpu = isNumeric(CoresUsed) ? Number(CoresUsed) * 1_000_000 : undefined;
+    const memory = isNumeric(MemoryUsed) ? Number(MemoryUsed) : undefined;
     const storage = isNumeric(StorageAllocatedSize) ? Number(StorageAllocatedSize) : undefined;
-    const cpuLimit = isNumeric(CoresLimit) ? Number(CoresLimit) : undefined;
+
+    const cpuUsage =
```
artemmufazalov, line 35: Can you please add a comment, why we currently use only `System` and `User` pools?

artemmufazalov (approved):

PR #549, opened 2023-10-11
feat: add ErrorBoundary

artemmufazalov:
Add `ErrorBoundary` to the app

Screen Shot 2024-01-12 at 10 15 34


src/containers/App/App.js
```diff
@@ -38,8 +39,10 @@ class App extends React.Component {
         const {singleClusterMode, clusterName} = this.props;
         return (
             <AsideNavigation>
-                <Content singleClusterMode={singleClusterMode} clusterName={clusterName} />
-                <div id="fullscreen-root"></div>
+                <ErrorBoundary>
+                    <Content singleClusterMode={singleClusterMode} clusterName={clusterName} />
+                    <div id="fullscreen-root"></div>
+                </ErrorBoundary>
```
artemmufazalov, line 45: The goal of boundary to allow user continue working in the app, so not everything is wrapped. 

So, on the one hand this doesn't cover possible problems with `Navigation`, `ThemeProvider` and `Authentication`. On the other hand, without them it would be just a page with some strange text, not really better, than the white screen we have currently

src/containers/App/Content.js
```diff
@@ -77,18 +80,20 @@ function ContentWrapper(props) {
...
+                            </Route>
+                            <Route>
+                                <ThemeProvider theme={theme}>
+                                    <div className={b({embedded: singleClusterMode})}>
+                                        {isAuthenticated ? props.children : <Authentication />}
+                                    </div>
+                                </ThemeProvider>
+                            </Route>
+                        </Switch>
+                    </ErrorBoundary>
```
artemmufazalov, line 96: First-level boundary to prevent completely white screen. It requires `Router`, so it doesn't cover the whole app, though it should be just enough

src/containers/App/App.js
artemmufazalov, line 45: Second-level boundary with navigation

Screen Shot 2023-10-11 at 12 15 09

src/utils/registerError.ts
```diff
@@ -0,0 +1,18 @@
+export function registerError(error: Error, message?: string, type = 'error') {
```
artemmufazalov, line 1: Decided not to add additional logic to pass rum logger from internal version since we already have all rum styles defined in embedded version

src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,58 @@
...
+                                {i18n('error-description')}
+                            </div>
+                            <Disclosure
+                                summary={i18n('show-details')}
+                                className={b('show-details')}
+                                size="m"
+                            >
+                                <pre className={b('error-details')}>{error.stack}</pre>
+                            </Disclosure>
+                            {onReportProblem && (
```
Raubzeug, line 41: Don't you want to add `Refresh` button? I've seen such a case in other projects.

src/containers/App/App.js
```diff
@@ -38,8 +39,10 @@ class App extends React.Component {
         const {singleClusterMode, clusterName} = this.props;
         return (
             <AsideNavigation>
-                <Content singleClusterMode={singleClusterMode} clusterName={clusterName} />
-                <div id="fullscreen-root"></div>
+                <ErrorBoundary>
```
Raubzeug, line 42: Is it enough only high-level EB? Maybe it will be usefull to use it also deeper?

Raubzeug (approved):

src/components/ErrorBoundary/ErrorBoundary.tsx
artemmufazalov: Good idea! Added

src/containers/App/App.js
artemmufazalov: It's first attempt, I'll add more low-level boundaries when it's will be clear where they are needed

Raubzeug (approved):

PR #551, opened 2023-10-11
fix(Tenant): fix tree not fully collapsed bug

artemmufazalov:
Fix the bug with not fully collapsed navigation tree. Steps:
1) Collapse navigation tree
2) Go to other page (Cluster, Node, Tablet, etc.)
3) Return to Tenant page

This happened because data was not updated properly

image

### DIFF
--- a/src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
+++ b/src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
@@ -58,14 +58,14 @@ import './ObjectSummary.scss';
 
 const b = cn('object-summary');
 
-const getInitialIsSummaryCollapsed = () => {
-    return Boolean(localStorage.getItem(DEFAULT_IS_TENANT_COMMON_INFO_COLLAPSED));
-};
+const getTenantCommonInfoState = () => {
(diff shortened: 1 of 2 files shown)


src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -58,16 +58,15 @@ import './ObjectSummary.scss';
...
-const initialTenantCommonInfoState = {
-    triggerExpand: false,
-    triggerCollapse: false,
-    collapsed: getInitialIsSummaryCollapsed(),
+    return {
+        triggerExpand: false,
+        triggerCollapse: false,
+        collapsed,
+    };
 };
```
artemmufazalov, line 69: When initial data was declared as const, it wasn't updated on component remount.
So, 
1. I somehow change setting value. Collapsed state is updated, initial `SplitPane` values update as well
2. I go to other page (`Cluster`, `Node`, `Tablet`, etc.)
3. `Tenant` component unmounts
4. I return to `Tenant` page
5. `Tenant` component mounts
6. Old initial data with not updated setting is taken for the reducer
7. However, `SplitPane` use update values. So we have not fully collapsed `SplitPane`, with minimal values

To fix this, reducer initial value recalculates on component mounts

Raubzeug (approved):

PR #564, opened 2023-10-19
feat: add top tables links

krosy1337:


src/containers/Tenant/Diagnostics/TenantOverview/TenantStorage/TopTables.tsx
```diff
@@ -68,11 +71,15 @@ export function TopTables({path}: TopTablesProps) {
             name: 'Path',
             sortable: false,
             render: ({row}) => (
-                <div className={b('cell-with-popover-wrapper')}>
+                <LinkToSchemaObject
+                    className={b('cell-with-popover-wrapper')}
+                    path={row.Path as string}
```
artemmufazalov, line 76: It's better to add some check before and pass `row.Path` without type assertion. Or at least have only `String(row.Path)`
```
if (!row.Path) {
   return null;
}        
```

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #565, opened 2023-10-19
fix: fix top queries table row height

krosy1337:


src/components/TruncatedQuery/TopQueriesTruncatedQuery.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const TopQueriesTruncatedQuery = ({value = ''}: TopQueriesTruncatedQueryProps) => {
+    const content = value.split('\n').slice(0, 5).join('\n');
+    return (
+        <CellWithPopover contentClassName={b('popover-content')} content={content}>
+            {value}
+        </CellWithPopover>
+    );
+};
```
artemmufazalov, line 20: Why only 5 lines is displayed in `Popover`?

Regarding `TruncatedQuery` and `TopQueriesTruncatedQuery`. I think they are two realisations of the same component, you could place them in 1 file. And you can call it something like `TruncatedQueryWithPopover`. Current naming is not good, because you have `TopQueries` in `Diagnostics`, but don't use your component in there. Second component should either be able to display data with defined `maxQueryHeight` (where 1 is special case with `white-space: nowrap;`) or have comment that it's one line query display. Or maybe it could even be named `OneLineQueryWithPopover`

Generally, it's good to name component scope-based, when they are part of some larger components and have defined place in component tree. And in this case they are placed near parent components. But for the more general-purpose components (we have them in `src/components` folder) it better to chose names, that reflects components purpose and functionality

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/TopQueries/getTopQueriesColumns.tsx
```diff
@@ -66,6 +86,20 @@ const userSIDColumn: Column<KeyValueRow> = {
     align: DataTable.LEFT,
 };
 
+const topQueriesQueryTextColumn: Column<KeyValueRow> = {
+    name: TOP_QUERIES_COLUMNS_IDS.TopQueriesQueryText,
+    header: tableColumnsNames[TOP_QUERIES_COLUMNS_IDS.TopQueriesQueryText],
+    render: ({row}) => <OneLineQueryWithPopover value={row.QueryText?.toString()} />,
+    sortable: false,
+};
```
artemmufazalov, line 94: Bad naming. You start name with "topQueries", but this column isn't used in initial `TopQueries` component, but in `TenantOverview`. It's better to think of better naming, probably not scope-based, but functionality-based

src/containers/Tenant/Diagnostics/TopQueries/getTopQueriesColumns.tsx
```diff
@@ -18,6 +23,21 @@ const TOP_QUERIES_COLUMNS_IDS = {
...
+
+const tableColumnsNames: Record<QueryColumns, string> = {
+    CPUTimeUs: 'CPUTimeUs',
+    QueryText: 'QueryText',
+    EndTime: 'EndTime',
+    ReadRows: 'ReadRows',
+    ReadBytes: 'ReadBytes',
+    UserSID: 'UserSID',
+    TopQueriesQueryText: 'QueryText',
+    QueryId: 'QueryId',
```
artemmufazalov, line 40: There is no much need to add columns headers to a variable, if there are not used somewhere else. For example, they can be used in Table columns setup control. But here there are used only for headers, so can be just placed in `header` field

artemmufazalov (changes requested):

artemmufazalov:
1. I think `QueryHash` will be a better naming. `QueryId` will intersects with queries in other our services, it supposes we have some queries ids stored somewhere, so there should be a functionality to track progress of long queries and retrieve results. `QueryId` isn't supposed to be calculated on a go and in UI

2. Negative ids looks weird for me

Screen Shot 2023-10-23 at 13 48 14

src/containers/Tenant/Diagnostics/TopQueries/getTopQueriesColumns.tsx
```diff
@@ -66,6 +72,23 @@ const userSIDColumn: Column<KeyValueRow> = {
...
+    render: ({row}) => <OneLineQueryWithPopover value={row.QueryText?.toString()} />,
+    sortable: false,
+};
+
+const queryHashColumn: Column<KeyValueRow> = {
+    name: TOP_QUERIES_COLUMNS_IDS.QueryHash,
+    render: ({row}) =>
+        // We use unsigned right shift operator (>>>) to avoid negative values
+        // eslint-disable-next-line no-bitwise
+        (crc32.str(String(row.QueryText)) >>> 0).toString(16).toUpperCase().padStart(8, '0'),
```
artemmufazalov, line 87: Can you please make it a separate function in `utils`? And could you please split it to steps and explain every step in comments
1. crc32
2. `>>>>`
3. toString(16)
4. toUpperCase()
5. padStart(8, '0')

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #566, opened 2023-10-19
fix(LinkToSchemaObject): fix schema link

krosy1337:


src/routes.ts
```diff
@@ -49,4 +49,9 @@ export function createHref(
     return `${compile(route)(params)}${search}`;
 }
 
+// embedded version could be located in some folder (e.g. host/some_folder/app_router_path)
+// window.location has the full pathname, while location from router ignores path to project
+export const getSchemaObjectPath = (query = {}) =>
```
artemmufazalov, line 54: I'd reconsider the naming. It's not only relates to schema objects, but it's a link, that does navigation not by the means of react router. Maybe `createExternalUILink`? And comment, that it's the navigation, that assumes page reload

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #567, opened 2023-10-23
fix: fix incorrect truncate strings with popover

krosy1337:
image

### DIFF
--- a/src/components/CellWithPopover/CellWithPopover.tsx
+++ b/src/components/CellWithPopover/CellWithPopover.tsx
@@ -17,8 +17,8 @@ export function CellWithPopover({
     ...props
 }: CellWithPopoverProps) {
     return (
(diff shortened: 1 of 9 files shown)


src/components/CellWithPopover/CellWithPopover.tsx
```diff
@@ -0,0 +1,26 @@
...
+    className,
+    wrapperClassName,
+    ...props
+}: CellWithPopoverProps) {
+    return (
+        <div className={b('wrapper', wrapperClassName)}>
+            <Popover className={b(null, className)} {...props}>
+                {children}
+            </Popover>
+        </div>
```
artemmufazalov, line 24: It better to put b-style higher. So you have `block__element` inside `block`. It will be more correct in terms of BEM methodology. More specific styles should be further in components tree

artemmufazalov (approved):

PR #568, opened 2023-10-23
fix: refactor metrics storage section

krosy1337:


src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.scss
```diff
@@ -118,4 +118,19 @@
...
+    &__cell-with-popover-wrapper {
+        display: flex;
+    }
+
+    &__cell-with-popover {
+        overflow: hidden;
+
+        white-space: nowrap;
+        text-overflow: ellipsis;
+    }
```
artemmufazalov, line 135: Can you please rebase this branch and replace this styles with `CellWithPopover` component?

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.scss
krosy1337: If we replace custom popover wrapper in `TopTables` with `CellWithPopover`, then we will have different color between link and `...`. If its ok I can use `CellWithPopover` there

image

artemmufazalov: We currently have such behaviour for `Nodes`. Is it possible to change it everywhere within `CellWithPopover`? If no, probably white dots will be fine

Screen Shot 2023-10-23 at 14 16 47

artemmufazalov (approved):

PR #571, opened 2023-10-24
fix: fix createExternalUILink

krosy1337:


src/routes.ts
```diff
@@ -25,6 +25,23 @@ export const parseQuery = (location: Location) => {
...
+    // if we pass raw route and there is a port in route, compile()
+    // will try to parse the port and throw an error
+    if (portMatch) {
+        const port = portMatch[0];
+        preparedRoute = route.replace(portRegExp, ':\\' + port.slice(1));
+    }
+
+    return preparedRoute;
+};
+
```
artemmufazalov, line 44: There are dynamic nodes with ports 31012, 31022, 31032 (it common case when they are used for some operations). Your code doesn't cover these cases. It's better to set 3-5 digits in row for the port regexp

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #582, opened 2023-11-13
fix: update links to VDisk and PDisk Developer UI

artemmufazalov:
Update links to PDIsk and VDisk DeveloperUI. They are no longer uses external logic and are always present. 

Links are formed based on current backend (node host with embedded-ui), they use clusters nodes ability to connect with each other. 

```{backend}/node/{tagetNodeId}/path```

Screen Shot 2023-11-13 at 18 02 31

### DIFF
--- a/src/containers/Node/NodeStructure/Pdisk.tsx
+++ b/src/containers/Node/NodeStructure/Pdisk.tsx
@@ -22,6 +22,7 @@ import InfoViewer, {type InfoViewerItem} from '../../../components/InfoViewer/In
 import {ProgressViewer} from '../../../components/ProgressViewer/ProgressViewer';
 import {Icon} from '../../../components/Icon';
 
+import i18n from '../i18n';
 import {Vdisk} from './Vdisk';
 import {valueIsDefined} from './NodeStructure';
 import {PDiskTitleBadge} from './PDiskTitleBadge';
@@ -91,6 +92,7 @@ function getColumns({
                                 className={b('external-button', {hidden: true})}
                                 href={vdiskInternalViewerLink}
                                 target="_blank"
+                                title={i18n('vdisk.developer-ui-button-title')}
                             >
                                 <Icon name="external" />
                             </Button>
@@ -230,6 +232,7 @@ export function PDisk({
                                 href={pDiskInternalViewerLink}
                                 target="_blank"
                                 view="flat-secondary"
+                                title={i18n('pdisk.developer-ui-button-title')}
                             >
                                 <Icon name="external" />
                             </Button>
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/utils/developerUI.ts [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+interface VDiskDeveloperUILinkParams extends PDiskDeveloperUILinkParams {
+    vDiskSlotId: number | string;
+}
+
+export const createVDiskDeveloperUILink = ({
+    nodeId,
+    pDiskId,
+    vDiskSlotId,
+}: VDiskDeveloperUILinkParams) => {
+    const vdiskPath = 'actors/vdisks/vdisk' + pad9(pDiskId) + '_' + pad9(vDiskSlotId);
```
Raubzeug, line 29: Maybe format string will be better?

src/utils/developerUI.ts
```diff
@@ -0,0 +1,32 @@
...
+    return `${backend}/node/${nodeId}/`;
+};
+
+interface PDiskDeveloperUILinkParams {
+    nodeId: number | string;
+    pDiskId: number | string;
+}
+
+export const createPDiskDeveloperUILink = ({nodeId, pDiskId}: PDiskDeveloperUILinkParams) => {
+    const pdiskPath = 'actors/pdisks/pdisk' + pad9(pDiskId);
```
Raubzeug, line 15: Maybe format string will be better?

src/containers/Node/NodeStructure/Pdisk.tsx
```diff
@@ -54,32 +55,36 @@ const vDiskTableColumnsNames: Record<VDiskTableColumnsIdsValues, string> = {
...
+                        vDiskSlotId,
+                    });
                 }
 
                 return (
                     <div className={b('vdisk-id', {selected: row.id === selectedVdiskId})}>
-                        <span>{row.VDiskSlotId}</span>
+                        <span>{vDiskSlotId}</span>
                         {vdiskInternalViewerLink && (
                             <Button
```
Raubzeug, line 89: Button here and in VDisk are without text. Lets add title to provide more accessibility.

Raubzeug (approved):

src/utils/developerUI.ts
artemmufazalov: As for me, current form is more explicit. I don't like formatted strings with symbols, because symbols are visually lost (`_` in this case)
```const vdiskPath = `actors/vdisks/vdisk${pad9(pDiskId)}_${pad9(vDiskSlotId)}`;```

Raubzeug (approved):

PR #584, opened 2023-11-14
fix(Nodes): always use nodes when flag is enabled

artemmufazalov:
1. Disable `useNodesEndpointInDiagnostics` flag by default
2. Explicitly set `storage=false` flag for nodes request (it's implicitly set by default)
3. Use `/nodes` instead of `/compute` for not DB entities in Diagnostics
4. Request entities nodes by `path` param instead of `tenant`

### DIFF
--- a/src/containers/Nodes/Nodes.tsx
+++ b/src/containers/Nodes/Nodes.tsx
@@ -5,7 +5,6 @@ import {useDispatch} from 'react-redux';
 import DataTable from '@gravity-ui/react-data-table';
 import {ASCENDING} from '@gravity-ui/react-data-table/build/esm/lib/constants';
 
-import type {EPathType} from '../../types/api/schema';
 import type {ProblemFilterValue} from '../../store/reducers/settings/types';
 import type {NodesSortParams} from '../../store/reducers/nodes/types';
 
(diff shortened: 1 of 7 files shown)


src/containers/Nodes/Nodes.tsx
```diff
@@ -90,15 +86,14 @@ export const Nodes = ({path, type, additionalNodesProps = {}}: NodesProps) => {
...
-            // since /nodes can return data only for tenants
-            if (path && (!useNodesEndpoint || !isDatabaseEntityType(type))) {
+            // If there is no path, it's cluster Nodes tab
+            if (path && !useNodesEndpoint) {
                 dispatch(getComputeNodes({path}));
             } else {
-                dispatch(getNodes({tenant: path}));
+                dispatch(getNodes({path}));
             }
         },
```
artemmufazalov, line 95: 1. Use `/nodes` instead of `/compute` for not DB entities in Diagnostics
2. Request entities nodes by `path` param instead of `tenant`

src/store/reducers/nodes/nodes.ts
```diff
@@ -99,9 +99,9 @@ const nodes: Reducer<NodesState, NodesAction> = (state = initialState, action) =
 };
 const concurrentId = 'getNodes';
 
-export function getNodes({type = 'any', ...params}: NodesApiRequestParams) {
+export function getNodes({type = 'any', storage = false, ...params}: NodesApiRequestParams) {
```
artemmufazalov, line 102: Explicitly set `storage=false` flag for nodes request (it's implicitly `false` by default)

Raubzeug (approved):

PR #585, opened 2023-11-15
fix(Tablets): reduce rerenders

artemmufazalov:


src/containers/Tablets/Tablets.tsx
```diff
@@ -34,14 +34,9 @@ interface TabletsProps {
 export const Tablets = ({path, nodeId, className}: TabletsProps) => {
     const dispatch = useDispatch();
 
-    const {
-        data = {},
```
artemmufazalov, line 38: `data` was used in `useMemo` dependencies. Because of this empty object assign, component infinitely rerendered until `data` was loaded

Raubzeug (approved):

PR #586, opened 2023-11-15
chore(Node): fix warnings

artemmufazalov:


src/containers/Node/NodeStructure/NodeStructure.tsx
```diff
@@ -62,7 +62,7 @@ function NodeStructure({nodeId, className}: NodeStructureProps) {
                 });
             }
         };
-    }, []);
+    }, [scrollContainer]);
```
Raubzeug, line 65: It is useless and misleading to set `ref.current` as `useEffect` dependency.

Raubzeug (changes requested):

src/containers/Node/NodeStructure/NodeStructure.tsx
artemmufazalov: Could you please clarify the issue here? I understand, why `ref` is useless, as it doesn't change. But `ref.current` could be updated and hence could trigger hooks

Raubzeug: The problem is that ref.current is not updated until component is rendered: here is an article in details: https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780

artemmufazalov: Thanks, I didn't know about this issue. It took me long to process the article, this behaviour isn't obvious at all

src/containers/Node/NodeStructure/NodeStructure.tsx
```diff
@@ -47,23 +47,11 @@ function NodeStructure({nodeId, className}: NodeStructureProps) {
...
-        return () => {
-            if (scrollContainer) {
-                scrollContainer.scrollTo({
-                    behavior: 'smooth',
-                    top: 0,
-                });
-            }
-        };
-    }, [scrollContainer]);
-
```
artemmufazalov, line 66: It occurred, this hook was never applied. Deleted it, as I don't understand, why it may be needed

src/containers/Node/NodeStructure/NodeStructure.tsx
```diff
@@ -77,13 +65,13 @@ function NodeStructure({nodeId, className}: NodeStructureProps) {
     }, [nodeId, dispatch]);
 
     useEffect(() => {
-        if (!isEmpty(nodeStructure) && scrollContainer) {
+        if (!isEmpty(nodeStructure) && scrollContainerRef.current) {
```
artemmufazalov, line 68: Returned old behaviour, where `scrollContainerRef.current` wasn't in dependency array. This case is the same, but doesn't cause linter warnings

Raubzeug (approved):

PR #590, opened 2023-11-16
fix(QueryResultTable): fix table error on null cell sort

artemmufazalov:
`viewer/json/query` returns `null` for empty table cells.

If type of table value was 'object', it was converted to string. But `typeof null === 'object'`.

Check `if (value === null)` didn't work because `null` was a string. However, when passed to `BigInt`, it was further converted that caused an error, table was stuck in invalid state.

Because of this issue sorting in `Preview` table could lead to `SyntaxError` (displayed on the screen). 

To fix it, I prevent converting `null` to string, when data is initially parsed.

Screen Shot 2023-11-16 at 17 11 28


src/components/QueryResultTable/QueryResultTable.tsx
```diff
@@ -36,10 +36,14 @@ const prepareTypedColumns = (columns: ColumnType[]) => {
...
-                if (value === undefined || value === null) return null;
+
+                if (value === undefined || value === null) {
+                    return null;
+                }
+
                 return columnType === 'number' ? BigInt(value) : value;
             },
-            render: ({value}) => <Cell value={value as string} />,
+            render: ({row}) => <Cell value={String(row[name])} />,
```
artemmufazalov, line 46: `value` type is always `unknown`. So I extract value from typed `row` value

Raubzeug (approved):

PR #593, opened 2023-11-21
feat(TenantOverview): add columns to memory table

artemmufazalov:
Add columns `Tablet Cache`, `Query Runtime` and `Sessions`

Example 1 (version before tablet cache and sessions were added to response)
Screen Shot 2023-11-21 at 17 07 32

Example 2 (current version, but with small usage)
Screen Shot 2023-11-21 at 17 08 19

### DIFF
--- a/src/containers/Nodes/getNodesColumns.tsx
+++ b/src/containers/Nodes/getNodesColumns.tsx
@@ -28,6 +28,9 @@ const NODES_COLUMNS_IDS = {
     Tablets: 'Tablets',
     TopNodesLoadAverage: 'TopNodesLoadAverage',
     TopNodesMemory: 'TopNodesMemory',
+    SharedCacheUsage: 'SharedCacheUsage',
+    MemoryUsedInAlloc: 'MemoryUsedInAlloc',
+    TotalSessions: 'TotalSessions',
 };
 
 interface GetNodesColumnsProps {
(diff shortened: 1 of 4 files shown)


src/types/api/nodes.ts
```diff
@@ -62,6 +65,20 @@ export interface TSystemStateInfo {
...
+    /**
+     * int64
+     *
+     * a positive value means the peer is ahead in time; a negative value means it's behind
+     */
+    MaxClockSkewWithPeerUs?: string;
+    MaxClockSkewPeerId?: number;
+
+    /** uint64 */
+    DisconnectTime?: string;
```
artemmufazalov, line 78: Values that are new in backend proto file. They received from backend, but are not used in current PR. Added to `interface` to keep it relevant

src/containers/Nodes/getNodesColumns.tsx
```diff
@@ -184,23 +187,61 @@ const topNodesLoadAverageColumn: NodesColumn = {
...
-    render: ({row}) =>
-        row.MemoryUsed ? (
-            <ProgressViewer
-                value={row.MemoryUsed}
-                capacity={row.MemoryLimit}
-                formatValues={formatStorageValuesToGb}
-                colorizeProgress={true}
-            />
-        ) : (
-            '—'
```
artemmufazalov, line 197: No data progress bar is more explicit than dash and suits more to neighbour cells and columns.
Screen Shot 2023-11-21 at 16 59 07

Raubzeug (approved):

PR #596, opened 2023-11-28
fix(Schema): display keys in right order

artemmufazalov:


src/containers/Tenant/Schema/SchemaViewer/SchemaViewer.tsx
```diff
@@ -75,22 +90,14 @@ export const SchemaViewer = ({keyColumnIds = [], columns = [], type}: SchemaView
...
     }
 
-    // Display key columns first
-    const tableData = columns.sort((column) => {
-        if (column.Id && keyColumnIds.includes(column.Id)) {
-            return 1;
-        }
-        return -1;
-    });
-
```
artemmufazalov, line 85: `initialSortOrder` should be used instead. Without sorting columns should be displayed as they are in `/json/describe` response

Raubzeug (approved):

PR #602, opened 2023-12-06
feat(Diagnostics): remove tenant diagnostics cards setting

artemmufazalov:
Always display metrics cards for DB diagnostics

Screen Shot 2023-12-06 at 16 49 58


src/containers/Tenant/Diagnostics/TenantOverview/TenantOverview.scss
```diff
@@ -33,57 +33,6 @@
         gap: 10px;
     }
 
-    &__common-info {
```
artemmufazalov, line 36: These styles was used in `OldTenantOverview`

PR #603, opened 2023-12-07
feat: always use localStorage if no settingsApi

artemmufazalov:
If no `settingsApi` is provided, app should always use `localStorage`

### DIFF
--- a/src/store/reducers/settings/settings.ts
+++ b/src/store/reducers/settings/settings.ts
@@ -127,7 +127,9 @@ export const setSettingValue = (
         if (settingsApi) {
             window.api.postSetting(name, value);
         } else {
-            localStorage.setItem(name, value);
+            try {
+                localStorage.setItem(name, value);
+            } catch {}
         }
     };
 };
--- a/src/store/reducers/settings/settings.ts
+++ b/src/store/reducers/settings/settings.ts
@@ -21,7 +21,12 @@ import {
 import '../../../services/api';
 import {parseJson} from '../../../utils/utils';
 import {QUERY_ACTIONS, QUERY_MODES} from '../../../utils/query';
-import {readSavedSettingsValue, systemSettings, userSettings} from '../../../utils/settings';
+import {
+    readSavedSettingsValue,
+    settingsApi,
+    systemSettings,
+    userSettings,
+} from '../../../utils/settings';
 
 import {TENANT_PAGES_IDS} from '../tenant/constants';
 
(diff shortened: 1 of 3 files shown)


src/store/reducers/settings/settings.ts
```diff
@@ -115,13 +120,14 @@ export const setSettingValue = (
...
-        const {singleClusterMode} = getState();
-        if (singleClusterMode) {
-            localStorage.setItem(name, value);
-        } else {
+
+        // If there is no settingsApi, use localStorage
+        if (settingsApi) {
             window.api.postSetting(name, value);
+        } else {
+            localStorage.setItem(name, value);
```
Raubzeug, line 130: Potentially this operation may fail, sp lets wrap it with `try ... catch`.

src/utils/settings.ts
```diff
@@ -3,8 +3,11 @@ import {getValueFromLS} from './utils';
 export const userSettings = window.userSettings || {};
 export const systemSettings = window.systemSettings || {};
 
+export const settingsApi = window.web_version ? systemSettings.settingsApi : undefined;
```
Raubzeug, line 6: What if we define settingsApi ourselves if it isn't defined initially? We know methods it should provide and can define it. Thus in other code there will be no need to worry, just use settingsApi.

Raubzeug (changes requested):

src/utils/settings.ts
artemmufazalov: Could you please clarify you question? I didn't get it

If you mean changing `settingsApi` on a go, this case isn't supported. The problem is that when we declare new `settingsApi` by ourselves, all currently stored settings should be copied there. Otherwise, there will be scenario, where our settings partly stored in `localStorage`, partly in other store, it will be a mess.

If you mean incapsulating this logic into some instance, that will provide certain API and decide where to store data within itself, I don't think it's really needed. The problem, that I solve here, is that we want to use `localStorage` both in internal and in embedded version. Now it will be a migration period, after which `settingsApi` will be deleted

Currently, it's first iteration to allow UI usage in Nebius, where there is no special backend service for user settings. After that I add some migration to code, that will copy data from `settingsApi` to `localStorage`. And as final step, use only `localStorage`

src/store/reducers/settings/settings.ts
artemmufazalov: What should be in `catch` section in this case?

src/utils/settings.ts
Raubzeug: Ok, I've got it! Lets stick to current version!

PR #641, opened 2024-01-13
docs: added initial roadmap version

antonkovalenko:
Would like to add public roadmap in a separate file
Currently embedded ui roadmap is in ydb/ROADMAP.md

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,33 @@
+# YDB Embedded UI Roadmap
+## Legend
+We use the following symbols as abbreviations:
+
+1. ㉕ - feature appeared in the Roadmap for 2025;
+1. ㉔ - feature appeared in the Roadmap for 2024;
+1. ✅ - feature has been released;
+1. 🚧 - feature is partially available and is under development;
+1. ❌ - feature has been refused;
+1. 🔥 - not yet released, but we are in rush.
+
+## Diagnostics
+
+1. 🚧 **Capacity diagnostics** – a set of signals on a database info tab to easily diagnose CPU or storage shortage
+1. 🚧 **Basic built-in charts** – support for charts display for most popular signals: resource consumption, requests and latency
+1. 🚧 **Display Table Hot Keys** – on a tab **Hot Keys** for a [column-oriented](https://ydb.tech/docs/en/concepts/datamodel/table#olap-data-types) and [row-oriented table](https://ydb.tech/docs/en/concepts/datamodel/table#row-orineted_table) display Hot Keys based on query statistics
+1. ㉔ **Network diagnostics** – display signals about cluster network performance and help to discover reasons for potential problems
+
+
+## Cluster Administration
+
+1. ㉔ **VDisk Eviction** – a special page for [VDisk](https://ydb.tech/docs/en/concepts/cluster/distributed_storage) with infos and support for [VDisk eviction](https://ydb.tech/docs/en/maintenance/manual/moving_vdisks) launch
+1. ㉔ **Display Configuration** – easy way to find out the value of any configuration paramater for a cluster or database
+1. ㉕ **Edit Configuration** – change the value of any configuration paramater for a cluster or database
+
+## Query and Schema Development
+
+1. 🚧 **Autocompletion** – autosuggestions to complete sql keywords and schema elements' names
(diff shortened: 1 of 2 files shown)


artemmufazalov:
We name commits according to conventional commits spec: https://www.conventionalcommits.org/en/v1.0.0/

You need to rename your commit and PR to something like `docs: added initial roadmap version" to pass the checks

ROADMAP.md
```diff
@@ -0,0 +1,33 @@
+# YDB Embedded UI Roadmap
+## Legend
+We use the following symbols as abbreviations:
+
+1. ㉕ - feature appeared in the Roadmap for 2025;
+1. ㉔ - feature appeared in the Roadmap for 2024;
+1. ✅ - feature has been released;
+1. 🚧 - feature is partially available and is under development;
```
artemmufazalov, line 8: Does this mean "in progress"? Or it mean feature exists, but works unstable or can be even turned in by feature-flag? Because currently you marked with this some of the points, that currently aren't added into UI - Autocompletion, Charts, Hot keys. Probably they should be 🔥, and the only point with 🚧 is Capacity diagnostics

ROADMAP.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Diagnostics
+
+1. 🚧 **Capacity diagnostics** – a set of signals on a database info tab to easily diagnose CPU or storage shortage
+1. 🚧 **Basic built-in charts** – support for charts display for most popular signals: resource consumption, requests and latency
+1. 🚧 **Display Table Hot Keys** – on a tab **Hot Keys** for a [column-oriented](https://ydb.tech/docs/en/concepts/datamodel/table#olap-data-types) and [row-oriented table](https://ydb.tech/docs/en/concepts/datamodel/table#row-orineted_table) display Hot Keys based on query statistics
+1. ㉔ **Network diagnostics** – display signals about cluster network performance and help to discover reasons for potential problems
+
+
+## Cluster Administration
```
artemmufazalov, line 20: Here we also discussed database creation and deletion, separate page for PDisk with support for PDisk restart

ROADMAP.md
```diff
@@ -0,0 +1,33 @@
...
+1. ㉔ **Display Configuration** – easy way to find out the value of any configuration paramater for a cluster or database
+1. ㉕ **Edit Configuration** – change the value of any configuration paramater for a cluster or database
+
+## Query and Schema Development
+
+1. 🚧 **Autocompletion** – autosuggestions to complete sql keywords and schema elements' names
+1. ㉔ **Support VIEWS** – basic support for new schema object type – `VIEW`
+1. ㉔ **Support Asyncronous Replication** – basic support for new feature – cluster-to-cluster asyncrounous replication 
+1. ㉔ **Display Keys and Column Families** – add wider schema tab in Diagnostics mode with information about column families, primary and partitioning keys.
+1. ㉕ **Change Table Column Width** – make it possible to alter a column (e.g. on a nodes page) width by dragging
```
artemmufazalov, line 32: It's not `Query and Schema Development`. Probably, some additional section like `Appearance` should be added

artemmufazalov (changes requested):

ROADMAP.md
antonkovalenko: >discussed database creation and deletion
I'm not sure we want this to be in the roadmap until we decided what to do with the feature
> separate pdisk page
I don't want to include features that don't add value per se. pdisk restart adds value. a new page with pdisk looks more like a UX reorg to me

artemmufazalov (approved):

ROADMAP.md
```diff
@@ -0,0 +1,38 @@
...
+1. ㉕ - feature appeared in the Roadmap for 2025;
+1. ㉔ - feature appeared in the Roadmap for 2024;
+1. ✅ - feature has been released;
+1. 🚧 - feature is partially available and is under development;
+1. ❌ - feature has been refused;
+1. 🔥 - not yet released, but we are in rush.
+
+## Diagnostics
+
+1. 🚧 **Capacity diagnostics** – a set of signals on a database info tab to easily diagnose CPU or storage shortage
```
fomichev3000, line 15: a year symbol is missed on this string and below

fomichev3000 (changes requested):

fomichev3000 (approved):

PR #648, opened 2024-01-15
feat: add clipboard button to nodes tree titles

vovaspace:


src/components/EntityStatus/EntityStatus.js
```diff
@@ -121,15 +119,13 @@ class EntityStatus extends React.Component {
                     {this.renderLink()}
                 </span>
                 {hasClipboardButton && (
-                    <Button
-                        component="span"
-                        size="s"
+                    <ClipboardButton
```
artemmufazalov, line 122: `ClipboardButton` doesn't disappear after copy event until some outside click

Screen Shot 2024-01-15 at 19 56 11

artemmufazalov, line 122: `ClipboardButton` in tables looks different compared to other buttons in extended version

Nodes table (old, with developer ui link)
Screen Shot 2024-01-15 at 19 51 53

Databases table (old, with monitoring link)
Screen Shot 2024-01-15 at 19 51 39

Databases table (new version, without additional buttons)
Screen Shot 2024-01-15 at 19 52 37

artemmufazalov (changes requested):

src/components/EntityStatus/EntityStatus.js
vovaspace: Refactored clipboard buttons a bit to visually unify them with Gravity UI.

vovaspace: Should looks similar now because both use Graviti's `Button`.

vovaspace: it's a feature, not a bug :)
The button stays focused after being clicked.

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * because `react-copy-to-clipboard` doesn't work with `Tooltip` otherwise.
+ */
+function InnerButton({
+    className,
+    status,
+    title,
+    ...props
+}: Omit<ClipboardButtonProps, 'text'> & {status: CopyToClipboardStatus}) {
+    return (
+        <Tooltip content={status === CopyToClipboardStatus.Success ? 'Copied!' : title || 'Copy'}>
```
artemmufazalov, line 29: Let's also add `placement`, probably 'right' will be the best option. Currently, when tooltip content changes, tooltip width changes as well, but not tooltip position. 

Before copy
Screen Shot 2024-01-16 at 17 38 58

After copy
Screen Shot 2024-01-16 at 17 38 48

artemmufazalov (changes requested):

src/components/ClipboardButton/ClipboardButton.tsx
vovaspace: Looks like a bug in Gravity. Fixed with `placement="bottom-start"`.

I don't like the placement on right because the tooltip overlaps the current table row.

Issue #655, opened 2024-01-19
Format chart values


artemmufazalov:
Format values, so they will be displayed on chart axes and chart tooltip with proper format:
1. Latency - ms
2. Size (memory, storage) - MB, GB, TB

<img width="898" alt="Screen Shot 2024-01-19 at 15 09 20" src="https://github.com/ydb-platform/ydb-embedded-ui/assets/67755036/9d8a7a56-bdd2-434a-b63e-6eb4be864a50">

Issue #640, opened 2024-01-12
Display requests chart at Database info tab


antonkovalenko:
Number of API requests per second

First part of https://github.com/ydb-platform/ydb-embedded-ui/issues/625

Issue #638, opened 2024-01-12
[charts] add timeframe selection control


antonkovalenko:
Timerange selection control must be displayed above all charts. Buttons like in Monitoring.

- 30m
- 1h
- 1d
- 1w

Issue #633, opened 2024-01-10
Add "untoggle" behaviour to diagnostics toggles


antonkovalenko:
Click on a toggled control that displays diagnostics info (CPU, Storage or Memory) must ontoggle it and display Info tab in its initial state

PR #657, opened 2024-01-22
feat(TenantOverview): add charts

artemmufazalov:
Should fix following issues:
https://github.com/ydb-platform/ydb-embedded-ui/issues/625
https://github.com/ydb-platform/ydb-embedded-ui/issues/633
https://github.com/ydb-platform/ydb-embedded-ui/issues/638
https://github.com/ydb-platform/ydb-embedded-ui/issues/640
https://github.com/ydb-platform/ydb-embedded-ui/issues/655


src/utils/versions/getVersionsColors.ts
```diff
@@ -9,6 +9,7 @@ export const hashCode = (s: string) => {
     }, 0);
 };
 
+// TODO: colors used in charts as well, need to move to constants
```
artemmufazalov, line 12: DIdn't move colors constants, so internal version imports won't be changed while it's being moved to open-source

src/utils/cn.ts
```diff
@@ -0,0 +1,3 @@
+import cn from 'bem-cn-lite';
+
+export {cn};
```
artemmufazalov, line 3: Make file with `{cn}` reexport due to two reasons:
1. Faster to import, since it named export, while lib export is default
2. Easier to update, if lib migration will be needed

Didn't update every usage, but will use it at least in new components.

src/containers/Tenant/Diagnostics/TenantOverview/MetricsCards/MetricsCards.tsx
```diff
@@ -72,22 +72,31 @@ export function MetricsCards({
 
     const queryParams = parseQuery(location);
 
+    // Allow tabs untoggle behaviour
+    const getTabIfNotActive = (tab: TenantMetricsTab) => {
```
artemmufazalov, line 76: Decided to preserve link behaviour, that we have for all other tabs

src/services/api.ts
```diff
@@ -377,7 +378,20 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
...
+    getChartData({target, from, until, maxDataPoints}: JsonRenderRequestParams) {
+        const requestString = `${target}&from=${from}&until=${until}&maxDataPoints=${maxDataPoints}&format=json`;
 
+        return this.post<JsonRenderResponse>(
+            this.getPath('/viewer/json/render'),
+            requestString,
+            {},
+            {
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded',
```
artemmufazalov, line 390: `viewer/json/render` requires different content type compared to other api calls

src/components/MetricChart/types.ts
```diff
@@ -0,0 +1,32 @@
+export type Metric =
+    | 'queries.requests'
+    | 'queries.latencies.p50'
+    | 'queries.latencies.p75'
+    | 'queries.latencies.p90'
+    | 'queries.latencies.p99'
+    | 'resources.cpu.usage'
+    | 'resources.memory.used_bytes'
+    | 'resources.storage.used_bytes';
```
artemmufazalov, line 9: Currently we have quite short list of metrics, so I enumerated them all here. Further, if more metrics will be added, some more complex types may be used: 
```
type Percentile = 'p50' | 'p75' | 'p90' | 'p99';
type QueryMetric = `queries.{string}.{Percentile}`;
type BytesMetric = `resources.{string}.used_bytes`
type Metric = QueryMetric | BytesMetric
```

src/store/reducers/tenant/tenant.ts
```diff
@@ -24,7 +24,7 @@ const SET_METRICS_TAB = 'tenant/SET_METRICS_TAB';
 const CLEAR_TENANT = 'tenant/CLEAR_TENANT';
 const SET_DATA_WAS_NOT_LOADED = 'tenant/SET_DATA_WAS_NOT_LOADED';
 
-const initialState = {loading: false, wasLoaded: false};
+const initialState = {loading: true, wasLoaded: false};
```
artemmufazalov, line 27: Tenant diagnostics tab content was requested twice, because requests were sent before state was set as loading and after tenant data is fully loaded. This fix seems to be the easiest, tenant content considered loading from the start, there is no attempt to load tab content

src/components/MetricChart/getDefaultDataFormatter.ts
```diff
@@ -0,0 +1,36 @@
...
+import type {ChartDataType, ChartValue} from './types';
+
+export const getDefaultDataFormatter = (dataType?: ChartDataType) => {
+    switch (dataType) {
+        case 'ms': {
+            return formatChartValueToMs;
+        }
+        case 'size': {
+            return formatChartValueToSize;
+        }
```
artemmufazalov, line 15: I covered only current scenarios, further more data types or data types sizes (for example `'Gb' | 'Tb' | 'auto'` for `size`) may be added

src/components/MetricChart/reducer.ts
```diff
@@ -0,0 +1,86 @@
...
+    wasLoaded: boolean;
+    data: PreparedMetricsData;
+    error: IResponseError | undefined;
+}
+
+export const initialChartState: ChartState = {
+    // Set chart initial state as loading, in order not to mount and unmount component in between requests
+    // as it leads to memory leak errors in console (not proper useEffect cleanups in chart component itself)
+    // TODO: possible fix (check needed): chart component is always present, but display: none for chart while loading
+    loading: true,
```
artemmufazalov, line 52: I want to make it with a separate component, to reuse it in Query and other component, that requires the same optimisation

adameat:
I think we'd discussed that before, I asked you to hide graphs completely if it's not supported on the database. I expect the very fast answer in that case.

artemmufazalov:
> I think we'd discussed that before, I asked you to hide graphs completely if it's not supported on the database. I expect the very fast answer from backend in that case.

1. Experiment in this case is not a big work, it easily added and easily removed if needed. I just don't want to add a feature to the UI, that is not complete. The fact, that data is present in that case doesn't guarantee its correctness, it's misleading. So, as for me, feature flag here is essential. When https://github.com/ydb-platform/ydb/issues/1065 is fixed, merged and released in some ydb tag, we could remove it
2. While charts are hidden by feature flag, it's not so essential to hide them if they don't work, so I'd implement this hide as a next step https://github.com/ydb-platform/ydb-embedded-ui/issues/659
3. As for the presented error view, I want to create some reusable chart component (I believe, the charts story will evolve event further). That's why error view is necessary. It is on the screenshot. But it doesn't mean, that disabled charts could not be hidden (but as I said, it's the next step)

adameat:
> > I think we'd discussed that before, I asked you to hide graphs completely if it's not supported on the database. I expect the very fast answer from backend in that case.
> 
> 1. Experiment in this case is not a big work, it easily added and easily removed if needed. I just don't want to add a feature to the UI, that is not complete. The fact, that data is present in that case doesn't guarantee its correctness, it's misleading. So, as for me, feature flag here is essential. When [[Viewer] Add proxy to database node for json_render request ydb#1065](https://github.com/ydb-platform/ydb/issues/1065) is fixed, merged and released in some ydb tag, we could remove it
> 2. While charts are hidden by feature flag, it's not so essential to hide them if they don't work, so I'd implement this hide as a next step [Hide diagnostics charts if they are not enabled #659](https://github.com/ydb-platform/ydb-embedded-ui/issues/659)
> 3. As for the presented error view, I want to create some reusable chart component (I believe, the charts story will evolve event further). That's why error view is necessary. It is on the screenshot. But it doesn't mean, that disabled charts could not be hidden (but as I said, it's the next step)

but you could skip whole feature-flag part... it would allow us to keep interface intact until there is actuall data for the graphs.

Raubzeug (approved):

sareyu:
when i switch btw day/week there is no much difference in charts, maybe because there is no data for the period
but i expect to see chosen timeframe with zero/empty line if no data available

sareyu (approved):

artemmufazalov:
> when i switch btw day/week there is no much difference in charts, maybe because there is no data for the period but i expect to see chosen timeframe with zero/empty line if no data available

Currently GraphShard doesn't return data for such timeframe (it's a limitation of the first version). It will be improved with the next versions. 

YAGR doesn't support ranges for timeline axis, so it could not be fixed easily. It could be fixed only by adding null pointes either on backend, or on frontend.

Discussed with @adameat, decided not to do anything (display only the data, that is received).

Created an issue, you could add your point there: https://github.com/ydb-platform/ydb-embedded-ui/issues/669

src/components/TimeFrameSelector/TimeFrameSelector.tsx
```diff
@@ -0,0 +1,33 @@
...
+    onChange: (value: TimeFrame) => void;
+    className?: string;
+}
+
+export const TimeFrameSelector = ({value, onChange, className}: TimeFrameSelectorProps) => {
+    return (
+        <div className={b(null, className)}>
+            {Object.keys(TIMEFRAMES).map((timeFrame) => {
+                return (
+                    <Button
```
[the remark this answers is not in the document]
artemmufazalov: I used yandex monitoring approach here (to have well recognised control), they display timeframes with buttons

src/containers/App/Content.js
```diff
@@ -80,18 +82,20 @@ function ContentWrapper(props) {
...
-                        </Route>
-                        <Route>
-                            <ThemeProvider theme={theme}>
-                                <div className={b({embedded: singleClusterMode})}>
-                                    {isAuthenticated ? props.children : <Authentication />}
-                                </div>
-                            </ThemeProvider>
-                        </Route>
-                    </Switch>
+                    <QueryParamProvider adapter={ReactRouter5Adapter}>
```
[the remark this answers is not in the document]
artemmufazalov: I plan. 

1. Currently somewhere we use some constructions like:
```
const {a = defaultA, b = defaultB, c} = parseQuery(location) as SomeType;
```
It works, but approach with a new hook in some cases seems more handy

2. Also we use [redux-location-state](https://github.com/spotify/redux-location-state) lib.
It's powerful, but it is not supported anymore. And I'm not sure, that I want to find an alternative for managing query params in redux state and middleware. It's not the approach, that I see in other our services. 

So, for me it looks like a good thing to try. Probably, we migrate to in-component search string management and get rid of redux-location-state. Probably, this solution won't be enough and further I'll try something else

artemmufazalov:
Further work will be with following issues:
https://github.com/ydb-platform/ydb-embedded-ui/issues/659
https://github.com/ydb-platform/ydb-embedded-ui/issues/669

PR #672, opened 2024-01-25
ci: add Embedded UI build to releases

vovaspace:
Also:
- refactor to stop using `yandex-cloud/ui-release-action` action;
- add `embedded_ui_refresh` event dispatch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,8 +8,43 @@ jobs:
   release:
     if: github.repository == 'ydb-platform/ydb-embedded-ui'
     runs-on: ubuntu-latest
+    env:
+      ASSET_NAME: embedded-ui
     steps:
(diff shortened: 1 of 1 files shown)


artemmufazalov:
Please add link to PR with ydb release action, when it's will be ready. 

As for me, ydb action should be merged first, so we won't trigger not existing workflow in this event (are there any errors in this case?)

vovaspace:
> Please add link to PR with ydb release action, when it's will be ready.

https://github.com/ydb-platform/ydb/pull/1321

> As for me, ydb action should be merged first, so we won't trigger not existing workflow in this event (are there any errors in this case?)

It seems to me that the trigger of an non-listening event should not cause an error. But this is only in theory :)

artemmufazalov (approved):

Issue #659, opened 2024-01-23
Hide diagnostics charts if they are not enabled


artemmufazalov:
Show charts only if backend returns some data

Charts may not work in following cases:
- Backend version, where charts are not supported - request will fall with some error (404, CORS, etc.)
- Feature flag with charts is not turned on

Also feature flag for charts should be removed. To turn on the charts endpoint (`viewer/json/render`), special Graph Shard should be configured and enabled. Until Graph Shard is enabled endpoint won’t work on a database. We assume, that if shard present, the data is correct and we could show the charts

PR #675, opened 2024-01-30
fix(TenantDashboard): hide if charts not enabled

artemmufazalov:
Issue: https://github.com/ydb-platform/ydb-embedded-ui/issues/659

Charts should be hidden, if they are not enabled. Since charts dashboard is quite big and causes layout shifts, it should be hidden under loader together with other content until charts data loads and it's clear, whether to show dashboard or not

Node, where charts are not enabled: https://nda.ya.ru/t/OnrxodIe74UwCV

Node, where charts enabled: https://nda.ya.ru/t/7w84LYSB74UwDE


src/components/LoadingContainer/LoadingContainer.tsx
```diff
@@ -0,0 +1,31 @@
...
+    loader?: ReactNode;
+    children?: ReactNode;
+    loading?: boolean;
+    className?: string;
+}
+
+// Use visibility:hidden to preserve loading content size
+// Also it helps to properly render charts and virtualized elements
+// that didn't render properly with display:none since they need parent container sizes
+export const LoadingContainer = ({
```
Raubzeug, line 19: I'm not sure that this additional wrapper above `Loader` is really essential. I would better stick to basic library component without wrappers.

src/containers/Tenant/Diagnostics/TenantOverview/TenantDashboard/TenantDashboard.tsx
```diff
@@ -26,16 +27,52 @@ export interface ChartConfig {
...
 }
 
-export const TenantDashboard = ({charts}: TenantDashboardProps) => {
+type TenantDashboardStatus = 'success' | 'error' | undefined;
+
+type DashboardsChartsStatuses = Record<string, ChartDataStatus>;
+
+export const TenantDashboard = ({charts, onDashboardLoad}: TenantDashboardProps) => {
+    const [dashboardStatus, setDashboardStatus] = useState<TenantDashboardStatus>();
+    const chartsStatuses = useRef<DashboardsChartsStatuses>({});
```
Raubzeug, line 39: It seem a bit redundant: you need to know if at least one chart is succeed. So, no need to check charts id's:
```
const [showDashboard, setShowDashboard] = useState(false);
... one chart was loaded => setShowDashboard(true) ...
... no chats was loaded => do nothing ...
if (!showDashboard) {
  return null;
}
```

Raubzeug (changes requested):

src/containers/Tenant/Diagnostics/TenantOverview/TenantCpu/TenantCpu.tsx
```diff
@@ -1,23 +1,33 @@
...
+
     return (
-        <>
-            <CpuDashboard />
+        <LoadingContainer loading={dashboardLoading} className={b('loading-container')}>
+            <TenantDashboard
+                charts={cpuDashboardConfig}
+                onDashboardLoad={() => setDashboardLoading(false)}
+            />
             <TopNodesByLoad path={path} additionalNodesProps={additionalNodesProps} />
```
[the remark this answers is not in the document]
artemmufazalov: `uplot` charts don't render correctly in components with `display:none`. Fixed this issue with a hack, removed additional containers

src/containers/Tenant/Diagnostics/TenantOverview/TenantDashboard/TenantDashboard.tsx
artemmufazalov: Yep, updated it. Didn't figured out in the beginning, that we can prevent charts refetches with autorefresh variable

src/containers/Tenant/Diagnostics/TenantOverview/TenantDashboard/TenantDashboard.tsx
```diff
@@ -29,39 +30,54 @@ interface TenantDashboardProps {
...
-    }
+    /**
+     * Charts should be hidden, if they are not enabled:
+     * 1. GraphShard is not enabled
+     * 2. ydb version does not have /viewer/json/render endpoint (400, 404, CORS error, etc.)
+     * If at least one chart successfully loaded, dashboard should be shown
+     * @link https://github.com/ydb-platform/ydb-embedded-ui/issues/659
+     * @todo disable only for specific errors ('GraphShard is not enabled') after ydb-stable-24 is generally used
+     */
+    const handleChartDataStatusChange = (chartStatus: ChartDataStatus) => {
```
Raubzeug, line 50: It seems that you are interested only in success finish. So, lets rename this to let's say `handleChartLoadingSucceeded` and execute it only in one case - if chart was successfully loaded.

src/utils/hooks/useIsElementVisible.ts
```diff
@@ -0,0 +1,38 @@
...
+        const el = ref.current;
+        if (el) {
+            observer.current?.observe(el);
+        }
+        return () => {
+            if (el) {
+                observer.current?.unobserve(el);
+            }
+        };
+    }, [ref]);
```
Raubzeug, line 35: `Ref` can't be a dependency.

src/components/MetricChart/MetricChart.tsx
```diff
@@ -112,7 +120,14 @@ export const MetricChart = ({
     width = 400,
     height = width / 1.5,
     chartOptions,
+    onChartDataStatusChange,
 }: DiagnosticsChartProps) => {
+    // YAGR charts use uplot inside
+    // uplot has the issue, that it doesn't render correctly inside not visible elements
+    // so if chart is used inside component with display:none, it will be empty, when visibility change
+    const containerRef = useRef<HTMLDivElement>(null);
+    const isChartVisible = useIsElementVisible(containerRef);
```
Raubzeug, line 129: It looks much more straightforward to pass `isChartVisible` property from parent.

Raubzeug (changes requested):

src/components/MetricChart/MetricChart.tsx
artemmufazalov: It's more straightforward, but it makes you to pass visible prop whenever you want to hide some component, that has charts inside. For example, if I'd want to add `NotRenderUntilFirstVisible` logic to tenant diagnostics tabs, it won't be possible without props drilling to ensure charts properly hidden and shown

src/containers/Tenant/Diagnostics/TenantOverview/TenantDashboard/TenantDashboard.tsx
artemmufazalov: I made quite a universal solution, to make it possible to extend it later - hide charts only on specific errors.

Currently they are hidden on every error, including timeouts when node is not healthy, some internal problems, etc. The reason - versions without render endpoint don't return specific error. 

Further charts should be hidden only on specific errors, when GraphShard is not enabled - added todo not to forger.

PR #676, opened 2024-01-30
fix(VirtualTable): optimise requests

artemmufazalov:
Updated `VirtualTable` could be seen here: https://nda.ya.ru/t/se1OYEka74Uxro

Ensure that experiment "Use table with data load on scroll for Nodes and Storage cluster tabs" is turned on before reviewing this table

### DIFF
--- a/src/components/VirtualTable/VirtualTable.tsx
+++ b/src/components/VirtualTable/VirtualTable.tsx
@@ -132,6 +132,16 @@ export const VirtualTable = <T,>({
         }
     }, []);
 
+    // Cancel all pending requests on component unmount
+    useEffect(() => {
+        return () => {
+            Object.values(pendingRequests.current).forEach((timer) => {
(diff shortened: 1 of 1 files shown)


src/components/VirtualTable/VirtualTable.tsx
```diff
@@ -76,7 +76,7 @@ export const VirtualTable = <T,>({
 
     const [error, setError] = useState<IResponseError>();
 
-    const [pendingRequests, setPendingRequests] = useState<Record<string, NodeJS.Timeout>>({});
+    const pendingRequests = useRef<Record<string, ReturnType<typeof setTimeout>>>({});
```
Raubzeug, line 79: You also need to cancel all `pendingRequest` when component unmounts.

Raubzeug (changes requested):

Raubzeug (approved):

PR #697, opened 2024-02-01
feat(VirtualTable): enable columns resize

artemmufazalov:
Add resize for "Version" column in Nodes table - little native resize triangle near right bottom corner.

Screen Shot 2024-02-02 at 11 10 41

The table could be reviewed in UI here: https://nda.ya.ru/t/aL9co3Y674WeyY

To enable `VirtualTable` turn on "Use table with data load on scroll for Nodes and Storage tabs " experiment in settings.

### DIFF
--- a/src/components/VirtualTable/TableHead.tsx
+++ b/src/components/VirtualTable/TableHead.tsx
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useRef, useState} from 'react';
+import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 
 import type {
     HandleTableColumnsResize,
@@ -134,15 +134,14 @@ export const TableHead = <T,>({
 }: TableHeadProps<T>) => {
     const [sortParams, setSortParams] = useState<SortParams>({});
 
-    const [resizeObserver, setResizeObserver] = useState<ResizeObserver | undefined>();
     const isTableResizeable = Boolean(onColumnsResize);
 
-    useEffect(() => {
+    const resizeObserver: ResizeObserver | undefined = useMemo(() => {
         if (!isTableResizeable) {
-            return;
+            return undefined;
         }
 
-        const newResizeObserver = new ResizeObserver((entries) => {
+        return new ResizeObserver((entries) => {
             const columnsWidth: TableColumnsWidthSetup = {};
             entries.forEach((entry) => {
                 // @ts-ignore ignore custrom property usage
@@ -152,8 +151,6 @@ export const TableHead = <T,>({
 
             onColumnsResize?.(columnsWidth);
         });
-
-        setResizeObserver(newResizeObserver);
     }, [onColumnsResize, isTableResizeable]);
 
     const handleCellMount = useCallback(
--- a/src/components/VirtualTable/TableHead.tsx
+++ b/src/components/VirtualTable/TableHead.tsx
@@ -132,14 +132,15 @@ export const TableHead = <T,>({
 }: TableHeadProps<T>) => {
     const [sortParams, setSortParams] = useState<SortParams>({});
 
-    const resizeObserver = useRef<ResizeObserver>();
+    const [resizeObserver, setResizeObserver] = useState<ResizeObserver | undefined>();
     const isTableResizeable = Boolean(onColumnsResize);
 
     useEffect(() => {
         if (!isTableResizeable) {
             return;
         }
-        resizeObserver.current = new ResizeObserver((entries) => {
+
+        const newResizeObserver = new ResizeObserver((entries) => {
             const columnsWidth: TableColumnsWidthSetup = {};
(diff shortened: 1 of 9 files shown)


src/containers/Nodes/getNodesColumns.tsx
```diff
@@ -89,6 +89,7 @@ const versionColumn: NodesColumn = {
         return <CellWithPopover content={row.Version}>{row.Version}</CellWithPopover>;
     },
     sortable: false,
+    resizeable: true,
```
artemmufazalov, line 92: Currently only `Version` column is resizeable, because proper resize also requires update for components used inside cells.

For example, `NodesHostWrapper` has fixed `330px` width , resize makes no sense for such component.

Other columns will be added further step by step, if needed

src/components/VirtualTable/TableHead.tsx
```diff
@@ -95,34 +195,19 @@ export const TableHead = <T,>({
...
-                                        defaultSortOrder={defaultSortOrder}
-                                    />
-                                </div>
-                            </th>
+                                column={column}
+                                sortOrder={sortOrder}
+                                defaultSortOrder={defaultSortOrder}
+                                onSort={handleSort}
+                                rowHeight={rowHeight}
+                                resizeObserver={resizeObserver.current}
```
Raubzeug, line 209: It seems it would be better not to pass `resizeObserver`, but make a component with `forwardRef` and do all observe staff here.

Raubzeug (changes requested):

sareyu:
image

add `min-width` for column and `overflow: ellipsys` for cells plz

sareyu (dismissed):

src/components/VirtualTable/TableHead.tsx
artemmufazalov: I intentionally created `TableHeadCell` component to `observe` and `unobserve` `<div>` on mount and unmount in `useEffect`. The reason - it's very important to correctly `unobserve` `<div>` on unmount, otherwise some odd values could be stored as column width.

In case with `forwardRef`, it's needed to store some reference to initial divs, to manually `unobserve` them on their unmount. So with removing `useEffect` construction from `HeadCell` some more complex construction will appear in `TableHead`

src/containers/Nodes/VirtualNodes.tsx
```diff
@@ -49,6 +50,8 @@ export const VirtualNodes = ({parentContainer, additionalNodesProps}: NodesProps
         NodesUptimeFilterValues.All,
     );
 
+    const [tableColumnsWidthSetup, setTableColumnsWidth] = useTableResize('nodesTableColumnsWidth');
```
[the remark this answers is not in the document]
artemmufazalov: I created it as a hook (not some prop inside `VirtualTable`) to reuse it with `react-data-table` (I assume `onTableResize` will be a callback from outside)

src/components/VirtualTable/TableHead.tsx
Raubzeug: Nonetheless, it's not correct to pass `ref.current` as property, because React doesn't guarantee to rerender children if `ref.current` will change.
So, as for me, array with refs is better solution. You may pass `unsubscribe` method to all `th`s.

Raubzeug (changes requested):

src/components/VirtualTable/TableHead.tsx
artemmufazalov: Always forget about this issue with `useRef`.

Replaced it with `useState`, now it should rerender properly.

As for the array of refs option, it won't be just array of refs, because there are cases, where columns could be unmounted one by one. Examples: tables with conditional columns - `Storage`, where degraded columns is shown only with specific filters setup, `Partitions`, where some columns appear, if consumer is selected; tables with columns setup - you can hide and show specific columns. So it either would be some additional events handlers, or some logic, to process this array to subscribe new columns and unsubscribe disappeared.

Raubzeug: What if we create functions in `TableHead` and pass it to all cells? Or this logic may be moved to custom hook.

```
const observedCells = React.useRef({})

const {observeCell, unobserveCell} = React.useMemo(() => {
  const resizeObserver = new ResizeObserver(......)

  const unobserveCell = (cellId) => {
  const cellRef = observedCells.current.cellId
  if (cellRef) {
    resizeObserver.unobserve(cellRef)
    delete observedCells.current.cellId
  }
}
const observeCell = (cellId, cellRef) => {
  unobserveCell(cellId)
  resizeObserver.unobserve(cellRef)
  observedCells.current.cellId = cellRef
}

  return {observeCell, unobserveCell}

}, [onColumnsResize, isTableResizeable])
```

Raubzeug (approved):

PR #710, opened 2024-02-09
feat: move cluster info to overview tab

### LINKED ISSUE #623: Move infos about cluster to a cluster tab
Currently there are several tabs on a cluster page:
* Databases
* Nodes
* Storage
* Version

Competitors have cluster info on a special tab.
It seemed to us more expected way to organize cluster parameters

Let's add a new tab **Overview** to the left of Databases
Move all the fields that are displayed currenty about cluster as they are

Remove special control and cluster info from the top section of the page
Leave only cluster name and HC status

ValeraS:
Closes #623

### DIFF
--- a/src/containers/Cluster/Cluster.tsx
+++ b/src/containers/Cluster/Cluster.tsx
@@ -1,5 +1,5 @@
 import {useEffect, useMemo, useRef} from 'react';
-import {useLocation, useRouteMatch} from 'react-router';
+import {Redirect, Switch, Route, useLocation, useRouteMatch} from 'react-router';
 import {useDispatch} from 'react-redux';
 import cn from 'bem-cn-lite';
 import qs from 'qs';
@@ -12,10 +12,10 @@ import type {
     AdditionalVersionsProps,
     AdditionalNodesProps,
 } from '../../types/additionalProps';
-import routes from '../../routes';
+import routes, {getLocationObjectFromHref} from '../../routes';
 
 import {setHeaderBreadcrumbs} from '../../store/reducers/header/header';
-import {getClusterInfo} from '../../store/reducers/cluster/cluster';
+import {getClusterInfo, updateDefaultClusterTab} from '../../store/reducers/cluster/cluster';
 import {getClusterNodes} from '../../store/reducers/clusterNodes/clusterNodes';
 import {parseNodesToVersionsValues, parseVersionsToVersionToColorMap} from '../../utils/versions';
 import {useAutofetcher, useTypedSelector} from '../../utils/hooks';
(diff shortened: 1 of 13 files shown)


src/containers/Cluster/Cluster.tsx
```diff
@@ -101,67 +102,127 @@ function Cluster({
...
+        activeTab = defaultTab;
+    }
+
+    useEffect(() => {
+        if (activeTab !== defaultTab) {
+            dispatch(updateDefaultClusterTab(activeTab));
+        }
+    }, [activeTab, defaultTab, dispatch]);
+
+    return {activeTab, defaultTab};
```
artemmufazalov, line 225: It looks like `activeTab` will be equal to `defaultTab`, since `defaultTab` is updated, if it differs

src/containers/Cluster/Cluster.tsx
```diff
@@ -101,67 +102,127 @@ function Cluster({
...
                                 {node}
                             </InternalLink>
                         );
                     }}
                 />
             </div>
 
-            <div>{renderTab()}</div>
+            <div>
+                <Switch>
```
artemmufazalov, line 147: What is the reason to use Router instead of switch case?

src/containers/Cluster/Cluster.tsx
```diff
@@ -101,67 +102,127 @@ function Cluster({
...
+                    </Route>
+                    <Route
+                        path={
+                            getLocationObjectFromHref(getClusterPath(clusterTabsIds.versions))
+                                .pathname
+                        }
+                    >
+                        <Versions versionToColor={versionToColor} />
+                    </Route>
+                    <Redirect to={getLocationObjectFromHref(getClusterPath(defaultTab))} />
```
artemmufazalov, line 200: Why do you need `Redirect` here? `activeTab` is always a `ClusterTab` and it equals to `defaultTab`, if it didn't matched before, something is wrong, `Redirect` won't help

artemmufazalov (changes requested):

src/containers/Cluster/Cluster.tsx
ValeraS: Yes, we update the default tab to the active tab because we want the next time we go to the cluster page to see that tab.

src/containers/Cluster/Cluster.tsx
ValeraS: It's easier here to redirect from `/cluster` to `/cluster/{defaultTab}` than to create proper references to the default tab throughout the codebase.

src/containers/Cluster/Cluster.tsx
ValeraS: we go to the cluster page (`/cluster`) and it redirects to the default tab.

src/containers/Cluster/Cluster.tsx
artemmufazalov: But why we need to return both `activeTab` and `defaultTab` from this hook it they are equal?

artemmufazalov (approved):

artemmufazalov (approved):

PR #713, opened 2024-02-12
fix: use DC and Rack from Location for nodes

artemmufazalov:
In raw data we have `DataCenter` field. However, in code we reference it as `DC`, also backend sort works only with field short name. So replaced `DataCenter` field with `DC` everywhere, use `DataCenter` and `Rack` from node `Location` field, where it's possible (it's not present in `/viewer/json/compute` endpoint)

`DC` and `Rack` should appear in Node page, in Nodes and Storage tables, also sort by DC in Nodes and Storage tables now should work properly (experiment "Use table with data load on scroll for Nodes and Storage tabs" should be turned off).

Node page:
Screen Shot 2024-02-12 at 15 51 40

Nodes table:
Screen Shot 2024-02-12 at 15 52 50

Node to test UI: https://nda.ya.ru/t/se1OYEka74Uxro

### DIFF
--- a/src/containers/Nodes/getNodesColumns.tsx
+++ b/src/containers/Nodes/getNodesColumns.tsx
@@ -14,6 +14,7 @@ import type {GetNodeRefFunc} from '../../types/additionalProps';
 import {getLoadSeverityForNode} from '../../store/reducers/nodes/utils';
 import {UsageLabel} from '../../components/UsageLabel/UsageLabel';
 import {CellWithPopover} from '../../components/CellWithPopover/CellWithPopover';
+import {EMPTY_DATA_PLACEHOLDER} from '../../utils/constants';
 
 const NODES_COLUMNS_IDS = {
     NodeId: 'NodeId',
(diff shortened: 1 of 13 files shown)


src/store/reducers/node/utils.ts
```diff
@@ -0,0 +1,16 @@
+import type {TEvSystemStateResponse} from '../../../types/api/systemState';
+import type {PreparedNode} from './types';
+
+export const prepareNodeData = (data: TEvSystemStateResponse): PreparedNode => {
```
artemmufazalov, line 4: Although is looks similar to other nodes preparation functions, it's different endpoint - `/viewer/json/sysinfo`, so there should be different handler

src/containers/Nodes/getNodesColumns.tsx
```diff
@@ -69,7 +70,7 @@ const dataCenterColumn: NodesColumn = {
     name: NODES_COLUMNS_IDS.DC,
     header: 'DC',
     align: DataTable.LEFT,
-    render: ({row}) => (row.DataCenter ? row.DataCenter : '—'),
+    render: ({row}) => row.DC || EMPTY_DATA_PLACEHOLDER,
```
artemmufazalov, line 73: In raw data we have `DataCenter` field. However, in code we reference it as `DC`, also backend sort works only with field short name. So replaced `DataCenter` field with `DC` everywhere, use `DataCenter` from node `Location` field, where it's possible (it's not present in `/viewer/json/compute` endpoint)

Raubzeug (approved):

PR #733, opened 2024-02-26
feat: show different page titles on different pages

### LINKED ISSUE #662: add cluster/db name in the start of ui page title
Image

As UI user I want to open several browser tabs with different ui pages to easily access different clusters/tables/dbs and multitask. I want to distinguish tabs by their titles without switching to them to work faster.

Currently all tabs have same title. It's impossible to find a tab I'm looking for without clickiong through all of them.

When in schema browsing context – let's output full path to the selected schema object in the beginning of the tab title.
When the database is known –  let's output full path to the database in the beginning of the tab title.
When the cluster is known, but database isn't let's output cluster name in the beginning of the tab title.

 I wouldn't output cluster name when the database is known because we use unique root for every cluster in most cases

ValeraS:
Closes #662


src/containers/Tablet/Tablet.tsx
```diff
@@ -98,57 +100,69 @@ export const Tablet = () => {
...
             </div>
-        </div>
+        );
+    };
+    return (
+        <>
+            <Helmet>
+                <title>{`${id} - ${
+                    tenantName || queryClusterName || i18n('tablet.header')
+                }`}</title>
```
artemmufazalov, line 163: For me `i18n('tablet.header')` is excessive here. It is not an alternative to tenant or cluster name if they are missing. For me it's more like `TabletId - YDB Monitoring` or `TabletId - Cluster - YDB Monitoring` in this case. Or if `TabletId` is not obvious indicator, something like `TabletId - Tablet - TenantName / ClusterName - YDB Monitoring` (`Tablet` is always present)

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #734, opened 2024-02-27
chore: update colors vars

artemmufazalov:
Somehow some of the colors weren't updated on uikit5 update

### DIFF
--- a/src/utils/versions/getVersionsColors.ts
+++ b/src/utils/versions/getVersionsColors.ts
@@ -25,7 +25,7 @@ export const COLORS = [
     '#b22222', // firebrick
 ];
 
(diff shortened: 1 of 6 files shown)


src/utils/versions/getVersionsColors.ts
```diff
@@ -25,7 +25,7 @@ export const COLORS = [
     '#b22222', // firebrick
 ];
 
-export const GRAY_COLOR = '#bfbfbf'; // --yc-color-base-neutral-hover
```
artemmufazalov, line 28: This color is no longer corresponds to uikit hover color, but it's left unchanged

Raubzeug (approved):

PR #739, opened 2024-02-29
feat: add binary data in plain text display

vovaspace:


src/containers/UserSettings/settings.ts
```diff
@@ -75,6 +76,16 @@ export const languageSetting: SettingProps = {
     },
 };
 
+export const binaryDataInPlainTextDisplay: SettingProps = {
+    settingKey: BINARY_DATA_IN_PLAIN_TEXT_DISPLAY,
+    title: i18n('settings.binaryDataInPlainTextDisplay.title'),
+    description:
```
vovaspace, line 82: My idea is to display description only in monitoring mode because in embedded mode this feature will only be provided with newer YDB versions.

Perhaps it would be better to move this condition somewhere, where is it better?

src/services/api.ts
```diff
@@ -40,6 +40,8 @@ import {backend as BACKEND, metaBackend as META_BACKEND} from '../store';
 import {prepareSortValue} from '../utils/filters';
 import {parseMetaCluster} from './parsers/parseMetaCluster';
 import {parseMetaTenants} from './parsers/parseMetaTenants';
+import {settingsManager} from './settings';
+import {BINARY_DATA_IN_PLAIN_TEXT_DISPLAY} from '../lib';
```
artemmufazalov, line 44: Could you please import not from '../lib' to prevent circular dependencies in the future?

src/containers/UserSettings/settings.ts
artemmufazalov: I think this condition is ok, I'd move it to some helper function `isMultiClusterVersion()`

src/containers/UserSettings/i18n/ru.json
```diff
@@ -14,6 +14,9 @@
   "settings.language.option-russian": "Русский",
   "settings.language.option-english": "English",
 
+  "settings.binaryDataInPlainTextDisplay.title": "Отображение двоичных данных в виде обычного текста",
```
artemmufazalov, line 17: Maybe we mention query requests here? Currently this text doesn't directly point on query requests

"Отображение двоичных данных в результатах запросов в виде обычного текста"

src/services/api.ts
```diff
@@ -305,9 +307,16 @@ export class YdbEmbeddedAPI extends AxiosWrapper {
         const uiTimeout = 9 * 60 * 1000;
         const backendTimeout = 10 * 60 * 1000;
 
+        const base64 = !settingsManager.readUserSettingsValue(
```
artemmufazalov, line 310: Could you please add comment explanation to `base64` param? Something like `encode string query results with base64`

artemmufazalov (changes requested):

src/containers/UserSettings/settings.ts
```diff
@@ -75,6 +76,16 @@ export const languageSetting: SettingProps = {
...
+        process.env.REACT_APP_META_BACKEND === undefined ||
+        process.env.REACT_APP_META_BACKEND === 'undefined'
```
ValeraS, line 84: Please do not use `REACT_APP_META_BACKEND` in common code. This is an env variable for our CRA. Let's make component like this
```
function Description() {
    const isSingleClusterMode = useTypedSelector((state) => state.singleClusterMode);
    if (isSingleClusterMode) {
        return null;
    }

    return <span>{i18n('settings.binaryDataInPlainTextDisplay.description')}</span>;
}

description: <Description />
```

ValeraS (changes requested):

src/containers/UserSettings/settings.ts
vovaspace: Moved to src/containers/ClusterModeGuard/ClusterModeGuard.tsx.

src/services/api.ts
vovaspace: Fixed.

src/containers/UserSettings/i18n/ru.json
vovaspace: Removed `ru` lang at all :)

src/containers/UserSettings/settings.ts
vovaspace: Added `src/containers/ClusterModeGuard/ClusterModeGuard.tsx` container for that.

src/services/api.ts
vovaspace: Yeah.

src/containers/UserSettings/i18n/ru.json
artemmufazalov: The same comment could be applied to the `en` version as well, it also lack "requests results"

src/containers/UserSettings/settings.ts
artemmufazalov: I think, name `AppModeGuard` will suit more here

vovaspace: I think base64 displaying should be the same throughout the application, not just in query. So I would make/name this setting very common.

src/containers/UserSettings/i18n/ru.json
vovaspace: I think base64 displaying should be the same throughout the application, not just in query. So I would make/name this setting very common.

src/containers/UserSettings/settings.ts
vovaspace: It's not clear to me what "app mode" means. I would assume that this is about prod/test envs.

artemmufazalov: You're right

artemmufazalov (approved):

ValeraS (changes requested):

ValeraS (dismissed):

ValeraS (approved):

PR #764, opened 2024-03-14
fix(MetricChart): draw area charts

artemmufazalov:
Make Grafana-styled area charts

Before:
Screen Shot 2024-03-18 at 18 18 54

After:
Screen Shot 2024-03-18 at 18 18 28

### DIFF
--- a/src/components/MetricChart/MetricChart.tsx
+++ b/src/components/MetricChart/MetricChart.tsx
@@ -1,13 +1,12 @@
 import {useCallback, useEffect, useReducer, useRef} from 'react';
 
-import {RawSerieData, YagrPlugin, YagrWidgetData} from '@gravity-ui/chartkit/yagr';
+import {YagrPlugin, YagrSeriesData, YagrWidgetData} from '@gravity-ui/chartkit/yagr';
 import ChartKit, {settings} from '@gravity-ui/chartkit';
 
 import type {IResponseError} from '../../types/api/error';
 import type {TimeFrame} from '../../utils/timeframes';
 import {useAutofetcher} from '../../utils/hooks';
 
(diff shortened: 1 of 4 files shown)


src/components/MetricChart/MetricChart.tsx
```diff
@@ -209,7 +216,7 @@ export const MetricChart = ({
                 dispatch(setChartError(err as IResponseError));
             }
         },
-        [metrics, timeFrame, width],
+        [database, metrics, timeFrame, width],
```
artemmufazalov, line 219: Forgot to add `database` here previously

src/components/MetricChart/colors.ts
```diff
@@ -0,0 +1,27 @@
...
+    '#EF843C', // 3: orange
+    '#E24D42', // 4: red
+    '#1F78C1', // 5: ocean
+    '#BA43A9', // 6: purple
+    '#705DA0', // 7: violet
+    '#508642', // 8: dark green
+    '#CCA300', // 9: dark sand
+];
+
+export function colorHexToRGBA(htmlColor: string, opacity: number) {
```
ValeraS, line 15: Maybe better to use special library for such manipulations. For example, https://www.npmjs.com/package/colord
With it, code will look like
```js
import {colord} from "colord";

export function colorToRgba(colorString: string; opacity: number) {
    const color = colord(colorString);
    if (!color.isValid()) {
        throw new Error(
            'Invalid color passed, the color should be in the html format. Example: #ff0033',
        );
    }
    
    return color.alpha(opacity).toRgbString();
}
```

artemmufazalov: Good idea, thanks!

PR #773, opened 2024-03-25
feat: add VDisk page

### LINKED ISSUE #617: VDisk page
All VDisk links should lead to this page.

Node Structure tab remains as it is.

On this page we should display VDisks info and info about related nodes and groups. Reference - VDisk info popup on Node Structure tab.

VDisk Developer UI link should be present.

Below VDIsk info there should be Storage table with groups / nodes radio button, filtered by VDisk id.

artemmufazalov:
Closes #617 

Screen Shot 2024-03-25 at 14 12 39


src/contexts/ClusterNodesMapContext/ClusterNodesMapContext.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {useTypedDispatch, useTypedSelector} from '../../utils/hooks';
+
+type ClusterNodesMapContextProps = {
+    children: ReactNode;
+};
+
+const ClusterNodesMapContext = createContext<NodesMap | undefined>(undefined);
+
+/** NodesMap - almost constant cluster nodes to hosts mapping */
+export function ClusterNodesMapContextProvider({children}: ClusterNodesMapContextProps) {
```
ValeraS, line 14: Why do we need context? Trigger `getNodesList` in the component that uses this provider and  loaded data will be in global store.

ValeraS (changes requested):

PR #776, opened 2024-03-28
fix: reduce initial bundle size

ValeraS:


src/utils/monaco/yqlSuggestions/generateSuggestions.ts
```diff
@@ -15,6 +15,39 @@ import {
     WindowFunctions,
 } from './constants';
 
+const CompletionItemKind: {
```
artemmufazalov, line 18: Why did you replace `monaco.languages.CompletionItemKind` with this object?

ValeraS: to remove dependency on `monaco-editor` and depend only on types.

artemmufazalov (approved):

PR #778, opened 2024-03-28
feat(formatNumber): render numbers with spaces instead of comas

kunga:


src/containers/App/App.scss
```diff
@@ -4,6 +4,9 @@
 * {
     // FIXME: this is an overkill, potentially could break external components, needs refactoring
     box-sizing: border-box;
+
+    // Make all numbers in the app monospace
+    font-variant-numeric: tabular-nums;
```
artemmufazalov, line 9: I deleted font for `TopShards` table. It's different from font throughout the app. Instead I set monospace numbers everywhere in the app

src/utils/bytesParsers/__test__/formatBytes.test.ts
```diff
@@ -9,23 +9,23 @@ describe('formatBytes', () => {
         expect(formatBytes({value: 100_000_000_000_000})).toBe('100 TB');
     });
     it('should convert to size', () => {
-        expect(formatBytes({value: 100_000, size: 'b'})).toBe('100,000 B');
-        expect(formatBytes({value: 100_000_000_000_000, size: 'gb'})).toBe('100,000 GB');
+        expect(formatBytes({value: 100_000, size: 'b'})).toBe('100 000 B');
```
artemmufazalov, line 12: Fixed broken tests

src/utils/numeral.ts
```diff
@@ -1,7 +1,14 @@
...
 
-import {i18n} from './i18n';
+import {Lang, i18n} from './i18n';
+
+// Set space delimeter for all locales possible in project
+Object.values(Lang).forEach((value) => {
+    if (numeral.locales[value]) {
+        numeral.locales[value].delimiters.thousands = ' ';
+    }
+});
```
artemmufazalov, line 11: You configured numeral almost correctly, but it seems you deleted "," in format string. In format "," is delimiter sign, not delimiter itself, so no delimiter was applied.

I made more simple delimiter override, so it should work and no so complex for future support

artemmufazalov:
Screen Shot 2024-03-29 at 15 27 23

kunga:
Cool thanks for the improvements 🔥

PR #791, opened 2024-04-03
fix(QueryEditor): use `full` stats mode instead of `profile`

vovaspace:


Raubzeug (commented):
Should the setting for enabling `stats=profile` be done here too?

P.S. discussed with @antonkovalenko , this will be done in #803

Raubzeug (approved):

PR #792, opened 2024-04-05
feat(TenantOverview): bars instead of circles

### LINKED ISSUE #673: Not obvious CPU card in database diagnostics
### Improve after UX-tests

1. Replace circle diagrams with horizontal bars
2. CPU – use only three (IC, User, System) thread pools
3. Display absolute consumption and limit on the bar, as it's done in other places. With units
4. Display red and yellow (and orange?) icons. Example below uses exclamation sign in a triangle. Every level must have a sign (except when it's okay)
5. Memory must be displayed using bar as well. I'm sure we will add more bars later for memory, but let's start with the one
6. Display switch title in hyperlink style so it would be obviously clickable

DB diagnostics CPU card shows max cpu usage among user and system pools. It's not obvious, metric "used cores / total db cores" is expected here

https://github.com/ydb-platform/ydb-embedded-ui/blob/main/src/store/reducers/tenants/utils.ts#L46

Screen Shot 2024-01-26 at 15 08 20

artemmufazalov:
Closes #673 

Screen Shot 2024-04-11 at 14 17 43

### DIFF
--- a/src/components/StatusIcon/StatusIcon.tsx
+++ b/src/components/StatusIcon/StatusIcon.tsx
@@ -5,7 +5,6 @@ import {cn} from '../../utils/cn';
 
 import CircleExclamationFillIcon from '@gravity-ui/icons/svgs/circle-exclamation-fill.svg';
 import CircleInfoFillIcon from '@gravity-ui/icons/svgs/circle-info-fill.svg';
-import CircleXmarkFillIcon from '@gravity-ui/icons/svgs/circle-xmark-fill.svg';
 import TriangleExclamationFillIcon from '@gravity-ui/icons/svgs/triangle-exclamation-fill.svg';
 
 import './StatusIcon.scss';
@@ -16,7 +15,7 @@ const icons = {
     [EFlag.Blue]: CircleInfoFillIcon,
     [EFlag.Yellow]: CircleExclamationFillIcon,
     [EFlag.Orange]: TriangleExclamationFillIcon,
-    [EFlag.Red]: CircleXmarkFillIcon,
+    [EFlag.Red]: CircleExclamationFillIcon,
 };
 
 export type StatusIconMode = 'color' | 'icons';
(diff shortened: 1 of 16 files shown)


src/store/reducers/tenants/utils.ts
```diff
@@ -23,21 +24,40 @@ const getTenantBackend = (tenant: TTenant) => {
...
+): TenantPoolsStats[] | undefined => {
     if (!poolsStats) {
         return undefined;
     }
 
-    const systemPoolUsage = poolsStats?.find(({Name}) => Name === 'System')?.Usage || 0;
-    const userPoolUsage = poolsStats?.find(({Name}) => Name === 'User')?.Usage || 0;
-    const icPoolUsage = poolsStats?.find(({Name}) => Name === 'IC')?.Usage || 0;
+    return poolsStats
+        .map<Partial<TenantPoolsStats | undefined>>((pool) => {
```
ValeraS, line 45:
```suggestion
        .map<TenantPoolsStats | undefined>((pool) => {
```

src/containers/Tenant/Diagnostics/TenantOverview/MetricsCards/MetricCard/MetricCard.scss
```diff
@@ -2,6 +2,9 @@
 
 .ydb-metrics-card {
     $b: &;
+
+    min-height: 252px;
+
     &__header {
```
ValeraS, line 8: let's add `gap: 8px`

ValeraS (dismissed):

src/components/StatusIcon/StatusIcon.tsx
```diff
@@ -16,7 +15,7 @@ const icons = {
     [EFlag.Blue]: CircleInfoFillIcon,
     [EFlag.Yellow]: CircleExclamationFillIcon,
     [EFlag.Orange]: TriangleExclamationFillIcon,
-    [EFlag.Red]: CircleXmarkFillIcon,
+    [EFlag.Red]: CircleExclamationFillIcon,
```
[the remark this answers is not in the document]
artemmufazalov: Because XMark icon gives wrong associations. We decided to change icons completely, but new set of icons from our designer is not ready yet. So the same icon but with different colors as temporary solution

ValeraS (approved):

PR #795, opened 2024-04-10
feat: use rtk query

ValeraS:


src/store/reducers/clusters/types.ts
```diff
@@ -33,10 +30,6 @@ export interface ClustersState extends ClustersFilters {
     list: PreparedCluster[];
 }
 
-export type ClustersAction =
-    | ApiRequestAction<typeof FETCH_CLUSTERS, PreparedCluster[], unknown>
-    | ReturnType<typeof changeClustersFilters>;
-
 export interface ClustersStateSlice {
-    clusters: ClustersState;
```
artemmufazalov, line 41: `ClustersState` type is unused and could be deleted

src/containers/Clusters/Clusters.scss
```diff
@@ -173,4 +173,8 @@
         display: flex;
         align-items: center;
     }
+
+    &__error {
+        @include mixins.body-2-typography();
+    }
```
artemmufazalov, line 179: Please, add `margin-left: 15px`, it'll look a little better

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #802, opened 2024-04-23
feat(Tenant): use rtk query

ValeraS:


src/containers/Tenant/Diagnostics/TenantOverview/TenantCpu/TopNodesByCpu.tsx
```diff
@@ -25,27 +15,18 @@ interface TopNodesByCpuProps {
...
-            dispatch(getTopNodesByCpu({tenant: path}));
-        },
-        [dispatch, path],
+    const {currentData, isFetching, error} = topNodesApi.useGetTopNodesQuery(
+        {tenant: path, sortValue: 'CPU'},
+        {pollingInterval: autorefresh ? DEFAULT_POLLING_INTERVAL : 0},
     );
 
-    useAutofetcher(fetchNodes, [fetchNodes], autorefresh);
+    const loading = isFetching && currentData === undefined;
```
Raubzeug, line 28: Here and further: isn't it possible to use `isLoading` property from query?

Raubzeug (approved):

src/containers/Tenant/Diagnostics/TenantOverview/TenantCpu/TopNodesByCpu.tsx
ValeraS: it repeats the current behaviour. `isLoading` is false when the query arguments change.

PR #814, opened 2024-05-13
Improve autocomplete

Raubzeug:


src/utils/monaco/yql/yql.keywords.ts
```diff
@@ -0,0 +1,13 @@
+export const keywords =
```
Raubzeug, line 1: This data should be moved to separate library. But for now I decided to copy it directly, because it significantly improve view of yql query in editor.

artemmufazalov (approved):

PR #816, opened 2024-05-13
test: add safari to playwright

artemmufazalov:


playwright.config.ts
```diff
@@ -22,6 +22,10 @@ const config: PlaywrightTestConfig = {
             name: 'chromium',
             use: {...devices['Desktop Chrome']},
         },
+        {
+            name: 'safari',
+            use: {...devices['Desktop Safari']},
```
Raubzeug, line 27: With Safari it could be useful to test special versions. Is it possible to do it?

Raubzeug (approved):

playwright.config.ts
artemmufazalov: Unfortunately, playwright uses the latests versions and does not allow custom versions.

"Playwright tests can only be executed on the browser versions that are shipped with the corresponding Playwright version"

As I understand, there is a workaround, where one can override custom browser path, but it looks not very reliable and easy to maintain since you need to install needed browser yourself

PR #823, opened 2024-05-15
feat: make tables resizeable

### LINKED ISSUE #807: Make tables resizeable

artemmufazalov:
Test stand: https://nda.ya.ru/t/xCSdBmdn75ueXj

All tables are resizeable. Resize values are stored to localStorage. Exception - tables without predefined set of columns (query result, preview, hotkeys), their resize values are stored only while component is mounted

Closes #807

### DIFF
--- a/src/components/DateRange/DateRange.scss
+++ b/src/components/DateRange/DateRange.scss
@@ -1,11 +1,18 @@
 .date-range {
     &__input {
         min-width: 190px;
+        height: 28px;
         padding: 5px 8px;
 
         color: var(--g-color-text-primary);
         border: 1px solid var(--g-color-line-generic);
         border-radius: var(--g-border-radius-m);
+        outline: none;
         background: transparent;
+    }
+
+    &__input:focus,
+    &__input:focus-visible {
+        border: 1px solid var(--g-color-line-generic-hover);
     }
 }
--- a/src/components/ResizeableDataTable/ResizeableDataTable.tsx
+++ b/src/components/ResizeableDataTable/ResizeableDataTable.tsx
@@ -1,7 +1,7 @@
 import type {DataTableProps, Settings} from '@gravity-ui/react-data-table';
 import DataTable, {updateColumnsWidth} from '@gravity-ui/react-data-table';
 
-import {cn} from '../..//utils/cn';
+import {cn} from '../../utils/cn';
 import {useTableResize} from '../../utils/hooks/useTableResize';
 
 import './ResizeableDataTable.scss';
--- a/src/components/ResizeableDataTable/ResizeableDataTable.tsx
+++ b/src/components/ResizeableDataTable/ResizeableDataTable.tsx
@@ -1,0 +1,43 @@
+import type {DataTableProps, Settings} from '@gravity-ui/react-data-table';
+import DataTable, {updateColumnsWidth} from '@gravity-ui/react-data-table';
+
+import {cn} from '../..//utils/cn';
+import {useTableResize} from '../../utils/hooks/useTableResize';
+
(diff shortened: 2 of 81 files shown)


src/components/DateRange/DateRange.scss
```diff
@@ -1,11 +1,18 @@
...
         border: 1px solid var(--g-color-line-generic);
         border-radius: var(--g-border-radius-m);
+        outline: none;
         background: transparent;
     }
+
+    &__input:focus,
+    &__input:focus-visible {
+        border: 1px solid var(--g-color-line-generic-hover);
+    }
```
artemmufazalov, line 17: These changes fix light outline, which doesn't correspond to dark mode and makes control height more, than others

Before:
Screen Shot 2024-05-15 at 17 09 31

After:
Screen Shot 2024-05-15 at 17 11 56

src/components/VirtualTable/ResizeHandler.tsx
```diff
@@ -0,0 +1,106 @@
...
+import {calculateColumnWidth, rafThrottle} from './utils';
+
+interface ResizeHandlerProps {
+    maxWidth?: number;
+    minWidth?: number;
+    getCurrentColumnWidth: () => number | undefined;
+    onResize?: (width: number) => void;
+}
+
+export function ResizeHandler({
```
artemmufazalov, line 13: Copied from `react-data-table` with some changes, since tables are different

src/components/ResizeableDataTable/ResizeableDataTable.tsx
```diff
@@ -0,0 +1,43 @@
+import type {DataTableProps, Settings} from '@gravity-ui/react-data-table';
+import DataTable, {updateColumnsWidth} from '@gravity-ui/react-data-table';
+
+import {cn} from '../..//utils/cn';
```
Raubzeug, line 4: typo in path

Raubzeug:
What do you think about making a bigger minWidth for columns, which contain tablets? For now I may shrink this columns for some reason, but it will be quite a tricky to enlarge it to previous size.

Raubzeug:
Is it possible to add action for double click on resize panel? It will be handy if double click restores initial column width.

Raubzeug:
Why didn't you make resizable columns with Vdisks and Pdisks in Storage page?

artemmufazalov:
> Is it possible to add action for double click on resize panel? It will be handy if double click restores initial column width.

It's possible, but it requires changes to `react-data-table`. Currently, we replace initial value with saved value, so initial value isn't accessible. 

I made an issue for this: https://github.com/ydb-platform/ydb-embedded-ui/issues/824

artemmufazalov:
> Why didn't you make resizable columns with Vdisks and Pdisks in Storage page?

They have their own styles with `overflow: visible`. Reasons:
1) Disks could have donors, donors displayed in stack, when expanded it's height bigger, than the row's height
2) There may be a lot of disks, up to 25 VDisks and 144 PDisks, they expand column width

We need some rework, if we want to change it. For now I decided to left it as it was. Also It's the last column, so its width isn't a big issue since it doesn't affect other columns and always have the width of the content

src/containers/ClusterModeGuard/ClusterModeGuard.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import {useTypedSelector} from '../../lib';
+import {useTypedSelector} from '../../utils/hooks';
```
artemmufazalov, line 3: Updated such imports as well to prevent circular deps when embedded-ui is used as package

Raubzeug (approved):

PR #830, opened 2024-05-20
fix: fix buttons incorrect align in Nodes and Tenants tables

### LINKED ISSUE #829: Controls do not align correctly in nodes table
Screen Shot 2024-05-17 at 14 46 59

artemmufazalov:
Fixes #829 

Before:
Screen Shot 2024-05-17 at 14 46 59

After:
Screen Shot 2024-05-17 at 18 07 17

### DIFF
--- a/src/components/EntityStatus/EntityStatus.scss
+++ b/src/components/EntityStatus/EntityStatus.scss
@@ -44,16 +44,6 @@
 
         white-space: nowrap;
         text-overflow: ellipsis;
-
-        a {
-            text-decoration: none;
-
-            color: var(--g-color-text-link);
-        }
-
-        a:hover {
-            color: var(--g-color-text-link-hover);
-        }
     }
 
     &__link_with-left-trim {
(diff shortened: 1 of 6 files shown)


src/components/EntityStatus/EntityStatus.scss
```diff
@@ -52,6 +44,16 @@
...
+
+        a {
+            text-decoration: none;
+
+            color: var(--g-color-text-link);
+        }
+
+        a:hover {
+            color: var(--g-color-text-link-hover);
+        }
```
artemmufazalov, line 56: Move styles inside `link` class, so link colors won't apply to buttons

Raubzeug: Why should we define these styles? Is it possible to use proper uikit components with needed styles?

Raubzeug (changes requested):

src/components/EntityStatus/EntityStatus.scss
artemmufazalov: React-router `Link` was used, it doesn't have proper styles. Replaced it with out link wrapper and removed unneeded styles

Raubzeug (approved):

PR #838, opened 2024-05-22
fix: fix copy button not hidden after copy

artemmufazalov:
Copy button didn't hide after copy. Made it be shown only on hover

Test stand: https://nda.ya.ru/t/01pCgQBZ763DPX

### DIFF
--- a/src/containers/App/App.scss
+++ b/src/containers/App/App.scss
@@ -121,7 +121,7 @@ body,
 .ydb-tree-view__item {
     &:hover {
         & .clipboard-button {
-            visibility: visible;
+            opacity: 1;
         }
     }
 }
--- a/src/containers/App/App.scss
+++ b/src/containers/App/App.scss
@@ -119,8 +119,7 @@ body,
 .data-table__row,
 .ydb-virtual-table__row,
 .ydb-tree-view__item {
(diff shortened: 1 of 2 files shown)


src/containers/App/App.scss
```diff
@@ -119,8 +119,7 @@ body,
 .data-table__row,
 .ydb-virtual-table__row,
 .ydb-tree-view__item {
-    &:hover,
-    &:focus-within {
+    &:hover {
```
Raubzeug, line 122: let's add `focus-visible` here, to make the button visible if iterating with keyboard.

Raubzeug (changes requested):

src/containers/App/App.scss
artemmufazalov: Elements with `visibility:hidden` cannot be selected focused with tab, so it doesn't have any effect

Raubzeug: 100%! What do you think if we hide them with `opacity=0`?

artemmufazalov: `opacity=0` works just fine, updated PR

Raubzeug (approved):

PR #839, opened 2024-05-23
feat(HealthCheck): show aggregated status in preview

ValeraS:


src/containers/Tenant/Diagnostics/TenantOverview/Healthcheck/HealthcheckPreview.tsx
```diff
@@ -45,17 +63,14 @@ export function HealthcheckPreview(props: HealthcheckPreviewProps) {
                         onClick={(event) => {
                             // FIXME: refactor card to remove the button from the anchor link.
                             event.preventDefault();
-                            onUpdate();
+                            refetch();
```
Raubzeug, line 66: If we refetch overview, but details are opened, it seems we should refetch them either.

Raubzeug (changes requested):

src/containers/Tenant/Diagnostics/TenantOverview/Healthcheck/HealthcheckPreview.tsx
Raubzeug: Discussed on standup, that when we open healthcheck details, we need also update healthcheck overview, so the data will be consistent.

src/containers/Tenant/Diagnostics/TenantOverview/Healthcheck/HealthcheckPreview.tsx
```diff
@@ -45,17 +63,14 @@ export function HealthcheckPreview(props: HealthcheckPreviewProps) {
...
                         }}
-                        loading={loading}
+                        loading={isFetching}
                         view="flat-secondary"
                     >
                         <Icon data={ArrowsRotateRight} size={20} />
```
Raubzeug, line 71: Could you please remove refresh button on card? Discussed to abandon it in favor of refresh control in toolbar.

Raubzeug (changes requested):

Raubzeug (approved):

PR #848, opened 2024-05-27
feat: disable controls for users with viewer rights

artemmufazalov:


src/containers/PDiskPage/PDiskPage.tsx
```diff
@@ -118,7 +119,7 @@ export function PDiskPage() {
                 <ButtonWithConfirmDialog
                     onConfirmAction={handleRestart}
                     onConfirmActionSuccess={handleAfterRestart}
-                    buttonDisabled={!nodeId || !pDiskId}
+                    buttonDisabled={!nodeId || !pDiskId || !isUserAllowedToMakeChanges}
```
Raubzeug, line 122: Here and farther I think we should add a tooltip with information why is this button disabled.

Raubzeug (changes requested):

src/containers/PDiskPage/PDiskPage.tsx
artemmufazalov: Added

Raubzeug (dismissed):

Raubzeug (approved):

PR #849, opened 2024-05-27
feat: add Nodes and Storage filters to url

ValeraS:


src/containers/Tenant/Diagnostics/Diagnostics.tsx
```diff
@@ -55,11 +56,12 @@ function Diagnostics(props: DiagnosticsProps) {
 
     const location = useLocation();
 
-    const queryParams = qs.parse(location.search, {
-        ignoreQueryPrefix: true,
-    });
+    const queryParams = new URLSearchParams(location.search);
+    for (const key of pagesQueryParams) {
+        queryParams.delete(key);
+    }
```
artemmufazalov, line 62: Why do you delete params here? Maybe there is another way to delete unneeded params? Further it may be needed to remove params for some other tab as well, If we will be removing params this way, it will be a mess

artemmufazalov (changes requested):

artemmufazalov:
Please, add saving params to url for `VirtualStorage` and `VirtualNodes` components (enabled with "Use table with data load on scroll for Nodes and Storage tabs" setting in experiments)

Corresponding issue: https://github.com/ydb-platform/ydb-embedded-ui/issues/818

artemmufazalov (approved):

PR #852, opened 2024-05-28
feat(Diagnostics): display tablets as table

ValeraS:


src/store/reducers/tablets.ts
```diff
@@ -43,3 +45,31 @@ export const tabletsApi = api.injectEndpoints({
...
+
+const getTabletsInfoSelector = createSelector(
+    (path: string) => path,
+    (path) => tabletsApi.endpoints.getTabletsInfo.select({path}),
+);
+
+const selectGetTabletsInfo = createSelector(
+    (state: RootState) => state,
+    (_state: RootState, path: string) => getTabletsInfoSelector(path),
+    (state, selectGetNodeStructure) => selectGetNodeStructure(state).data,
```
artemmufazalov, line 57: `selectGetNodeStructure` - wrong naming

src/store/reducers/tablets.ts
```diff
@@ -43,3 +45,31 @@ export const tabletsApi = api.injectEndpoints({
...
+
+export const selectTabletsWithFqdn = createSelector(
+    (state: RootState, path: string) => selectGetTabletsInfo(state, path),
+    (state: RootState) => selectNodesMap(state),
+    (data, nodesMap): (TTabletStateInfo & {fqdn?: string})[] => {
+        if (!data?.TabletStateInfo) {
+            return [];
+        }
+        if (!nodesMap) {
+            return data as any;
```
artemmufazalov, line 68: Why you return `any` here? It may be just `data.TabletStateInfo`, in case `viewer/json/nodelist` (`nodesMap`) will be responding too long, it will be just a table without fqdns. But in your case some problems with `nodelist` could break table

src/containers/Tenant/Diagnostics/Tablets/Tablets.tsx
```diff
@@ -0,0 +1,171 @@
...
+import i18n from './i18n';
+
+const b = cn('tablets-table');
+
+const columns: DataTableColumn<TTabletStateInfo & {fqdn?: string}>[] = [
+    {
+        name: 'Type',
+        get header() {
+            return i18n('Type');
+        },
```
artemmufazalov, line 33: Why you define header property this way and not just `header: i18n('Type')`?

src/containers/Tenant/Diagnostics/Tablets/Tablets.tsx
```diff
@@ -0,0 +1,171 @@
...
+    },
+    {
+        name: 'Uptime',
+        get header() {
+            return i18n('Uptime');
+        },
+        render: ({row}) => {
+            return calcUptime(row.ChangeTime);
+        },
+        sortAccessor: (row) => -Number(row.ChangeTime),
```
artemmufazalov, line 102: Please, add `align: 'right'` to `Uptime`, it makes easier row to row comparison, when values have different length (for example "16d 16:16:16" and "0:16:16")

src/containers/Tenant/Diagnostics/Tablets/Tablets.tsx
```diff
@@ -0,0 +1,171 @@
...
+    }
+    const {currentData, isFetching, error} = tabletsApi.useGetTabletsInfoQuery(params, {
+        pollingInterval: autorefresh,
+    });
+
+    const loading = isFetching && currentData === undefined;
+    const tablets = useTypedSelector((state) => selectTabletsWithFqdn(state, path || ''));
+
+    if (loading) {
+        return <Loader />;
```
artemmufazalov, line 155: `<TableSkeleton/>` will suit more here, since it's a table

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/Tablets/Tablets.tsx
ValeraS: Theoretically, i18n should be called inside components/actions.

artemmufazalov (approved):

PR #853, opened 2024-05-28
feat: delete icons sprite, use gravity icons

artemmufazalov:
1. In many places we used icons from sprite. It's not very handy
2. Since we used our own icons, they may differ from the icons in components from uikit, thus we may have several versions of the same icon on one page

Deleted sprite, replaced most of the icons with uikit versions

Test stand: https://nda.ya.ru/t/lQGqdSKK768yCA


Raubzeug:
Let's also use this library for illustrations? https://github.com/gravity-ui/illustrations

Raubzeug (approved):

PR #859, opened 2024-05-30
fix(Tenant): always set tenantPage in the url

ValeraS:


src/containers/AsideNavigation/useNavigationMenuItems.ts
```diff
@@ -26,8 +26,8 @@ export function useNavigationMenuItems() {
     const location = useLocation();
     const history = useHistory();
 
-    const [initialTenantPage, setInitialTenantPage] = useSetting<string>(TENANT_INITIAL_PAGE_KEY);
-    const {tenantPage = initialTenantPage} = useTypedSelector((state) => state.tenant);
+    const [, setInitialTenantPage] = useSetting<string>(TENANT_INITIAL_PAGE_KEY);
```
artemmufazalov, line 29: Case, where `tenantPage` is changed not via navigation buttons is uncovered. For example, page is changed when preview is triggered, on query click in `TopQueries` tab, etc. In these cases values isn't stored in LS

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #893, opened 2024-06-10
feat(Header): treat breadcrumbs as links

NaiSidera:
Also "leader" hints are removed (we assume a tablet is a leader by default now)

### DIFF
--- a/src/containers/Header/Header.tsx
+++ b/src/containers/Header/Header.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import type {BreadcrumbsItem} from '@gravity-ui/uikit';
 import {Breadcrumbs} from '@gravity-ui/uikit';
 import {useHistory, useLocation} from 'react-router';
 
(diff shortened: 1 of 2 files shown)


src/containers/Tablets/Tablets.tsx
```diff
@@ -32,11 +32,7 @@ const columns: DataTableColumn<TTabletStateInfo & {fqdn?: string}>[] = [
             return i18n('Type');
         },
         render: ({row}) => {
-            return (
-                <span>
-                    {row.Type} {row.Leader ? <Text color="secondary">leader</Text> : ''}
-                </span>
-            );
+            return <span>{row.Type}</span>;
```
artemmufazalov, line 35: Tablet isn't a leader by default. If `row.Leader === false`, there should be `follower` text. `leader` text is unneeded, because it's true in most of the cases

src/containers/Header/Header.tsx
```diff
@@ -63,12 +63,7 @@ function Header({mainPage}: HeaderProps) {
         const breadcrumbs = getBreadcrumbs(page, options, rawBreadcrumbs, queryParams);
 
         return breadcrumbs.map((item) => {
-            const action = () => {
-                if (item.link) {
-                    history.push(item.link);
-                }
-            };
-            return {...item, action};
+            return {...item, href: item.link};
```
artemmufazalov, line 66: It fully reloads page in this case. Can we make it `react-router` link, so the app will behave like SPA?

src/containers/Tablets/Tablets.tsx
Raubzeug, line 35: You should render `follower` if `!row.Leader`.

artemmufazalov (changes requested):

artemmufazalov:
Tablet `leader` text is a separate issue, please split such changes to separate PRs further

src/containers/Header/Header.tsx
Raubzeug, line 66: in this case we may use `renderItem` property.

Raubzeug (changes requested):

PR #896, opened 2024-06-10
chore: use browserslist config from gravity-ui

### LINKED ISSUE #653: Use `@gravity-ui/browserslist-config`
We use a lot of libs from `@gravity-ui`. It will be correct to use [@gravity-ui/browserslist-config](https://github.com/gravity-ui/browserslist-config) as well to have similar browser support

NaiSidera:
Closes #653

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,21 +112,13 @@
       "^@gravity-ui/uikit/toaster-singleton-react-18$": "@gravity-ui/uikit/build/cjs/toaster-singleton-react-18.js"
     }
   },
-  "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -110,9 +110,7 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
+      "extends @gravity-ui/browserslist-config"
```
Raubzeug, line 113: It seems there's no sense to split setting into production and development. Let's combine these parts into one.

Raubzeug (changes requested):

package.json
NaiSidera: Done!

Raubzeug (approved):

PR #920, opened 2024-06-14
fix: revert release, fix package

artemmufazalov:
It should fix ydb-embedded-ui npm package

### DIFF
--- a/src/lib.ts
+++ b/src/lib.ts
@@ -16,7 +16,7 @@ export {setUserSettings} from './store/reducers/settings/settings';
 export {componentsRegistry} from './components/ComponentsProvider/componentsRegistry';
 export {useSetting, useTypedSelector} from './utils/hooks';
 export {getMonitoringLink, getMonitoringClusterLink} from './utils/monitoring';
-export {i18n, Lang} from './utils/i18n';
+export {i18n, Lang, registerKeysets} from './utils/i18n';
 export {toaster} from './utils/createToast';
 export {cn} from './utils/cn';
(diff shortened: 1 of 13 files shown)


src/types/api/tenant.ts
```diff
@@ -128,13 +128,7 @@ export interface ControlPlane {
...
 }
 
-export enum ETenantType {
-    'UnknownTenantType' = 'UnknownTenantType',
-    'Domain' = 'Domain',
-    'Dedicated' = 'Dedicated',
-    'Shared' = 'Shared',
-    'Serverless' = 'Serverless',
-}
+export type ETenantType = 'UnknownTenantType' | 'Domain' | 'Dedicated' | 'Shared' | 'Serverless';
```
artemmufazalov, line 131: Converted enum to type, so there will be no need to import it

src/styles/constants.scss
artemmufazalov, line 1: Moved constants to `App.scss`, so there will be no need to import them separately

src/lib.ts
```diff
@@ -31,3 +31,4 @@ export type {
 } from './containers/UserSettings/settings';
 export type {SettingProps} from './containers/UserSettings/Setting';
 export type {AsideNavigationProps} from './containers/AsideNavigation/AsideNavigation';
+export type {GetMonitoringLink, GetMonitoringClusterLink} from './utils/monitoring';
```
artemmufazalov, line 34: Added missing exports

Raubzeug (approved):

Issue #918, opened 2024-06-14
Fix `undefined:undefined` in query issues


artemmufazalov:
![2024-06-14 13 32 30](https://github.com/ydb-platform/ydb-embedded-ui/assets/67755036/71900eb7-7197-4080-843a-b579495405e7)

PR #932, opened 2024-06-18
fix: handle with undefined value of row and column in issue position,…

walborn:
1. handled with undefined value of row and column in issue position
2. removed 'file' heritage


Raubzeug (approved):

walborn:
closes #918

PR #944, opened 2024-06-25
feat(QueryEditor): save query on Cmd/Ctrl + S

NaiSidera:


ValeraS (changes requested):

src/containers/Tenant/Query/QueryEditor/keybindings.ts
```diff
@@ -0,0 +1,19 @@
+import type Monaco from 'monaco-editor';
+
+import {userOs} from '../../../../utils/getUserOS';
+
+export function getKeyBindings(monaco: typeof Monaco) {
+    const {KeyMod, KeyCode} = monaco;
+
+    const ctrlKey = userOs === 'mac' ? KeyMod.CtrlCmd : KeyMod.WinCtrl;
```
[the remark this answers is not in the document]
Raubzeug: Yes, that's it.

ValeraS (changes requested):

src/utils/keyboard.ts
```diff
@@ -0,0 +1,83 @@
+interface GetUserOS {
```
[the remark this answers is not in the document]
Raubzeug: It seems we do... We use shortcuts as button titles, so it should be strings, not components. What do think if we add such a utility to uikit?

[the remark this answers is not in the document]
ValeraS: We can use ActionTooltip instead of native title prop. Let's think about it in another PR. It's not relevant to this one.

src/containers/Tenant/Query/QueryEditor/keybindings.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    const ctrlKey = KeyMod.CtrlCmd;
+
+    /* eslint-disable no-bitwise */
+    return {
+        sendQuery: ctrlKey | KeyCode.Enter,
+        sendSelectedQuery: ctrlKey | KeyMod.Shift | monaco.KeyCode.Enter,
+        selectPreviousQuery: ctrlKey | KeyCode.UpArrow,
+        selectNextQuery: ctrlKey | KeyCode.DownArrow,
+        saveQuery: ctrlKey | KeyCode.KeyS,
```
artemmufazalov, line 14: It conflicts with default system or browser save key. It saves current page instead of query

src/containers/Tenant/Query/SaveQuery/SaveQuery.tsx
```diff
@@ -0,0 +1,151 @@
...
+        <Dialog
+            open={queryAction === 'save'}
+            hasCloseButton={false}
+            size="s"
+            onClose={onCloseDialog}
+            onEnterKeyDown={onSaveClick}
+        >
+            <Dialog.Header caption={i18n('action.save')} />
+            <Dialog.Body className={b('dialog-body')}>
+                {singleClusterMode && <div className={b('dialog-row')}>{i18n('description')}</div>}
```
artemmufazalov, line 120: This string is no longer depends on cluster mode, we always save in `localStorage`

artemmufazalov (changes requested):

src/containers/Tenant/Query/QueryEditor/keybindings.ts
Raubzeug: If editor is focused, it will save current query. If not, page will be saved. Please try it!

artemmufazalov: I tried. For me it saves the page, not the query

artemmufazalov: It works. Maybe I did something wrong

artemmufazalov (dismissed):

ValeraS (approved):

PR #955, opened 2024-06-27
feat(Acl): improve view

### LINKED ISSUE #939: Improve ACL
1) Access Type. In 99.99999% it is "Allow". Let's check if another value is presented, and add this column only if it does.
2) Access Rights. If backend return high-level rights, it will be handy to show only them (maybe with tooltip, where full list of rights is shown?)
3) Subject. Should be visible (without ellipsis). If backend return type of subject, it may be useful to show icons (user, robot, group, etc) to visually identify what subject is it.
4) Inheritance Type. It takes a lot of space. Maybe we can show icons for every type?

5) Maybe we should abandon table in favor of cards or definition list?

Raubzeug:
closes #939


src/containers/Tenant/Acl/Acl.tsx
```diff
@@ -80,14 +67,34 @@ export const Acl = ({path}: {path: string}) => {
...
+
+        const items = normalizedAcl.map(({Subject, ...data}) => {
+            return {
+                label: Subject,
+                items: aclParams
+                    .map((key) => {
+                        const value = data[key];
+                        if (!value) {
+                            return undefined;
+                        }
```
artemmufazalov, line 80: It looks like `Access` is alway an array. 

Screenshot 2024-07-02 at 11 45 07

Database: https://nda.ya.ru/t/aTbam5AN76jBrf

artemmufazalov, line 80: How it should look like for the user with default rights? Just `Subject`?

artemmufazalov (changes requested):

artemmufazalov (dismissed):

artemmufazalov (approved):

artemmufazalov (dismissed):

artemmufazalov (approved):

PR #969, opened 2024-07-02
fix(TenantNavigation): fix radio button value

### LINKED ISSUE #965: Tenant selected tab isn't changed by actions
`Query | Diagnostics`

If you click on buttons, tab is changed together with buttons value. But if the tab is changed via some other actions, for example by selecting template in schema tree, or by clicking on query in `Top Queries`, buttons values isn't changed

artemmufazalov:
Closes #965

### DIFF
--- a/src/containers/Tenant/TenantNavigation/TenantNavigation.tsx
+++ b/src/containers/Tenant/TenantNavigation/TenantNavigation.tsx
@@ -40,7 +40,7 @@ export const TenantNavigation = () => {
                 onUpdate={handleUpdate}
                 size="l"
                 className={b('body')}
-                defaultValue={getCurrentItem().id}
+                value={getCurrentItem().id}
                 options={navigationItems.map(transformItemToOption)}
             />
         </div>


src/containers/Tenant/TenantNavigation/TenantNavigation.tsx
```diff
@@ -31,16 +34,14 @@ export const TenantNavigation = () => {
...
-
     return (
         <div className={b()}>
             <RadioButton
                 width="auto"
                 onUpdate={handleUpdate}
                 size="l"
                 className={b('body')}
-                defaultValue={getCurrentItem().id}
+                value={tenantPage}
```
Raubzeug, line 44: It seems that it will be enough just change property from `defaultValue` to `value`. 
Lets stick to this and return all tabs calculations to `useTenantNavigation`?

Raubzeug (changes requested):

src/containers/Tenant/TenantNavigation/TenantNavigation.tsx
artemmufazalov: I don't understand, why we need to set `current` property to `items` and get current item with `getCurrentItem`, if we could just use `tenantPage`. For me it just adds complexity

PR #975, opened 2024-07-03
fix: update deps

artemmufazalov:


package.json
```diff
@@ -122,52 +122,52 @@
...
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
   },
   "overrides": {
     "fork-ts-checker-webpack-plugin": "^9.0.2",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
+    "react": "^18.3.1",
```
Raubzeug, line 169: If we need to use our current version in overrides, we may use the next syntax:
`"react": "$react"`

Raubzeug (changes requested):

Raubzeug (approved):

Issue #914, opened 2024-06-13
autorefresh improvements


antonkovalenko:
1. [ ] Do not clear contents on error, display error in popup at right bottom corner
2. [x] Add autorefresh control to every page that does refresh

PR #976, opened 2024-07-03
feat(AutoRefresh): use auto refresh control on other pages

ValeraS:
#914


src/containers/PDiskPage/PDiskPage.tsx
```diff
@@ -128,6 +129,7 @@ export function PDiskPage() {
                     <Icon data={ArrowRotateLeft} />
                     {pDiskPageKeyset('restart-pdisk-button')}
                 </ButtonWithConfirmDialog>
+                <AutoRefreshControl />
```
artemmufazalov, line 132: It's located too close to restart button. As for me, it creates association, that it somehow correlates with `PDIsk` restart, not data refecth

In my view, this control should not be on the same level as restart button. 

The same comment applicable to `VDisk` and it's evict button.

Screenshot 2024-07-04 at 11 55 46

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #977, opened 2024-07-03
chore: update deps

ValeraS:


artemmufazalov:
Please, merge this commit with version update ("Release-as" or "fix" prefix) and probably different name (e.g. "package update"). Since it changes how files are located in package, it should not be just "chore"

artemmufazalov (approved):

PR #995, opened 2024-07-09
 fix: rename VirtualTable to PaginatedTable, remove old hooks

artemmufazalov:


src/containers/Storage/Storage.tsx
```diff
@@ -95,25 +89,15 @@ export const Storage = ({additionalNodesProps, tenant, nodeId}: StorageProps) =>
...
 
-    const nodesRequestParams = useNodesRequestParams({
-        filter,
-        nodesUptimeFilter: uptimeFilter,
-        ...nodesSortParams,
-    });
-    const storageRequestParams = useStorageRequestParams({
-        filter,
-        ...groupsSortParams,
-    });
```
artemmufazalov, line 106: Outdated hooks, in current implementation they always return `undefined`. They depended on old param `USE_BACKEND_PARAMS_FOR_TABLES_KEY` which was replaced with `USE_PAGINATED_TABLES_KEY`, so if `USE_PAGINATED_TABLES_KEY === false`, hooks return undefined, if `USE_PAGINATED_TABLES_KEY === true`, `PaginatedStorage` is used

Raubzeug (approved):

PR #1005, opened 2024-07-10
fix: store types should be stored separately

Raubzeug:


src/types/store/queryActions.ts
artemmufazalov, line 1: Can you instead create `queryActions` folder inside `/reducers` and put both reducer and types there?

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #1063, opened 2024-07-24
feat: use monaco-yql-languages for syntax highlight

Raubzeug:
[Stand](https://nda.ya.ru/t/CFCL_PWV77AJ7a)


src/containers/Tenant/Query/QueryEditor/QueryEditor.tsx
```diff
@@ -375,7 +374,7 @@ function QueryEditor(props: QueryEditorProps) {
                     <div className={b('monaco-wrapper')}>
                         <div className={b('monaco')}>
                             <MonacoEditor
-                                language={LANGUAGE_YQL_ID}
+                                language={'yql'}
```
artemmufazalov, line 377: Can you use `LANGUAGE_ID` variable here, since you use it everywhere else?

src/containers/Tenant/Query/ExplainResult/ExplainResult.js
```diff
@@ -141,7 +141,7 @@ export function ExplainResult(props) {
         const content = (
             <div className={b('ast')}>
                 <MonacoEditor
-                    language={LANGUAGE_S_EXPRESSION_ID}
+                    language={LANGUAGE_ID}
```
artemmufazalov, line 144: It doesn't work for ast

Before:
Screenshot 2024-07-25 at 14 09 50


Now:
Screenshot 2024-07-25 at 14 05 40

artemmufazalov (changes requested):

src/containers/Tenant/Query/QueryEditor/QueryEditor.tsx
Raubzeug: Unfortunately no. It seems that editor is initializing before `monaco-yql-languages` is ready. So, it's not available. And i decided to use a hardcode here, instead of another app-level constant to avoid possible clinches.

artemmufazalov (approved):

Issue #872, opened 2024-06-05
Use pdisk/info endpoint for PDisk page


artemmufazalov:
Based on https://github.com/ydb-platform/ydb/pull/5195

adameat:
![image](https://github.com/ydb-platform/ydb-embedded-ui/assets/34044711/6b3315d7-b758-44eb-808c-8ae12ef8edcf)


pdisk properties:
* node_id - pdisk_id (I expect it to be in the header)
* somewhere on the top, where Restart button is - DecomissionStatus (DecommitStatus)
* BSC/Type, Path, Guid, SerialNumber
* BSC/StatusV2, State, Device, Realtime
* Space: AvailableSize / TotalSize (as one progress bar)
* Slots: BSC/NumActiveSlots / BSC/ExpectedSlotCount (as one progress bar)
* Log: LogUsedSize/LogTotalSize (as one progress bar)  SystemSize (it would be nice to show it on one row)

it would be also nice to show BSC/SharedWithOs attribute as a label. it's only important when it's true. so, it could be absent otherwise.

PR #1069, opened 2024-07-25
feat(PDiskPage): add pdisk attributes, display in 2 columns

artemmufazalov:
Closes: #872 

Screenshot 2024-07-26 at 00 55 42

### DIFF
--- a/src/components/PDiskInfo/PDiskInfo.tsx
+++ b/src/components/PDiskInfo/PDiskInfo.tsx
@@ -5,6 +5,7 @@ import {cn} from '../../utils/cn';
 import {formatStorageValuesToGb} from '../../utils/dataFormatters/dataFormatters';
 import {createPDiskDeveloperUILink} from '../../utils/developerUI/developerUI';
 import type {PreparedPDisk} from '../../utils/disks/types';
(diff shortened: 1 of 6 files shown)


src/components/PDiskInfo/PDiskInfo.tsx
```diff
@@ -107,14 +169,49 @@ export function PDiskInfo<T extends PreparedPDisk>({
...
+        pDisk,
+        nodeId,
+        isPDiskPage,
+        isUserAllowedToMakeChanges,
+    });
+
+    return (
+        <div className={b('wrapper', className)}>
+            <div className={b('col')}>
+                <InfoViewer info={generalInfo} renderEmptyState={() => null} />
```
Raubzeug, line 208: What do you think if we start using `DefinitionList` from `@gravity-ui/components`?

Raubzeug (approved):

src/components/PDiskInfo/PDiskInfo.tsx
artemmufazalov: I try to use it for new components (e.g. `ViewInfo`), but don't refactor old components (maybe someday I'll do it separately, currently I don't want to mix features with refactoring)

PR #1083, opened 2024-07-29
feat(ExecuteResult): show query schema with stats

ValeraS:


src/components/Graph/Graph.tsx
artemmufazalov, line 1: It looks like you create component that is similar to the component created during refactoring here: https://github.com/ydb-platform/ydb-embedded-ui/pull/1061

Was it intentionally? If no, I advise somehow discuss it with @Raubzeug, either separate refactoring from https://github.com/ydb-platform/ydb-embedded-ui/pull/1061 and do your changes based on it or wait until PR https://github.com/ydb-platform/ydb-embedded-ui/pull/1061 is finished  (it seems it almost ready) to prevent future merge conflics or code duplications

artemmufazalov (changes requested):

Raubzeug (dismissed):

src/components/Graph/Graph.tsx
Raubzeug: Let's use Valera's implementation! I'll rebase after merge this PR.

src/components/Graph/Graph.tsx
```diff
@@ -0,0 +1,57 @@
...
+        graphRoot.innerHTML = '';
+
+        const topology = getTopology(graphRoot.id, data, opts, shapes);
+        topology.render();
+        return () => {
+            topology.destroy();
+        };
+    }, [data, opts, shapes]);
+
+    return <div id={containerId} ref={containerRef} style={{height: '100vh'}} />;
```
artemmufazalov, line 33: There is much more info now, `100vh` isn't enough for expanded schema. Could we make `height` dynamic?

Screenshot 2024-07-29 at 18 00 31

artemmufazalov (changes requested):

src/components/Graph/Graph.tsx
ValeraS: I don't see an easy way to do this. 100vh was chosen as a height greater than the maximum height of the container. The same should be done with the width. If schema does not fit into visible area, then you can move it with the mouse

artemmufazalov: I didn't know, that it can be moved by mouse. In this case, it's not a big problem

artemmufazalov (approved):

Issue #1052, opened 2024-07-22
Infinite describe requests if describe returns `null`


artemmufazalov:
When some schema object is bugged, and describe returns `null` for it, it causes infinite loader on Describe tab and infinite requests on Overview tab

PR #1086, opened 2024-07-29
fix: infinite describe requests

ValeraS:
if describe is finished with error or empty result


src/containers/Tenant/Tenant.tsx
```diff
@@ -78,10 +78,7 @@ export function Tenant(props: TenantProps) {
     const {PathType: currentPathType, PathSubType: currentPathSubType} =
         currentItem?.PathDescription?.Self || {};
 
-    let showBlockingError = false;
-    if (error && typeof error === 'object' && 'status' in error) {
-        showBlockingError = error.status === 403;
-    }
+    const showBlockingError = Boolean(error && !currentItem);
```
artemmufazalov, line 81: `null` response could be for schema objects with problems, it doesn't mean, that the whole page is broken. For example, when I created issue https://github.com/ydb-platform/ydb-embedded-ui/issues/1052 I have the database, where I have two columns tables, that have `null` in schema response (sorry I didn't mentioned that). At the same tame, all other 100+ schema objects was just fine.

Now we have `PageError` even if one object is broken and I try to access it on first load (if I access it as result of navigation, it just displays "Schema is not available", it's that's needed

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
```diff
@@ -92,11 +92,14 @@ function Overview({type, path, database}: OverviewProps) {
...
-        return <ResponseError error={schemaError || overviewError} />;
-    }
-
-    return renderContent();
+    return (
+        <React.Fragment>
+            {schemaError || overviewError ? (
+                <ResponseError error={schemaError || overviewError} />
+            ) : null}
+            {renderContent()}
```
artemmufazalov, line 100: There are two errors now - one from error here, and one from component, that displays content

Screenshot 2024-07-30 at 11 13 29

artemmufazalov (changes requested):

src/containers/Tenant/Diagnostics/Overview/Overview.tsx
ValeraS: Yes, there can be two messages here. One request may fail with an error (or we interpret empty data as an error), while another may fail with no error but no data.

artemmufazalov (approved):

PR #1091, opened 2024-07-30
fix(Nodes): better memory bars

kunga:
## Changes

- Render memory without a limit as full
- Use bars for memory in Nodes list
- Render `Caches` instead of `Tablet Caches` and `Query Runtime`
- Render values with `formatStorageValuesToGb` instead of `formatBytesToGigabyte` for better precision

#### Before

image

#### After

image

#### Before

image

#### After

image

### DIFF
--- a/src/components/ProgressViewer/ProgressViewer.tsx
+++ b/src/components/ProgressViewer/ProgressViewer.tsx
@@ -86,6 +86,10 @@ export function ProgressViewer({
         } else if (fillWidth > dangerThreshold) {
             status = inverseColorize ? 'good' : 'danger';
         }
+        if (!isNumeric(capacity)) {
+            status = 'good';
+            fillWidth = 100;
+        }
     }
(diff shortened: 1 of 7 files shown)


src/components/ProgressViewer/ProgressViewer.tsx
```diff
@@ -86,6 +86,10 @@ export function ProgressViewer({
         } else if (fillWidth > dangerThreshold) {
             status = inverseColorize ? 'good' : 'danger';
         }
+        if (!isNumeric(capacity)) {
+            status = 'good';
```
artemmufazalov, line 90: Please, do not set status here. In nodes memory case it will be good anyway, but this will break the case with `inverseColorize=true`

artemmufazalov (changes requested):

artemmufazalov (approved):

Issue #1032, opened 2024-07-18
option to hide domain


adameat:
add option "hide root domain on database list". set it to true by default.

when we get databases list, and it length more than 1 and it has elements with Type != "Domain" we remove all elements with Type == "Domain" (we expect only 1 element removed)

PR #1094, opened 2024-07-30
feat: option to hide domain

astandrik:
Closes https://github.com/ydb-platform/ydb-embedded-ui/issues/1032


src/containers/UserSettings/i18n/en.json
```diff
@@ -37,6 +37,9 @@
   "settings.usePaginatedTables.title": "Use paginated tables",
   "settings.usePaginatedTables.description": " Use table with data load on scroll for Nodes and Storage tabs. It will increase performance, but could work unstable",
 
+  "settings.useHideDomainDatabase.title": "Use hide domain database",
+  "settings.useHideDomainDatabase.description": "Hide root domain on database list",
```
artemmufazalov, line 41: root = domain, it's better only leave title like "Hide domain database". And better call it "Show domain database", since it's disabled by default

src/containers/UserSettings/settings.tsx
```diff
@@ -167,6 +174,7 @@ export const experimentsSection: SettingsSection = {
     settings: [
         useNodesEndpointSetting,
         usePaginatedTables,
+        useHideDomainDatabase,
```
artemmufazalov, line 177: Should it be in experiments? It's not actual experiment. Maybe introduce some section with advanced settings, that help you to customise UI (no idea, how it could be called)?

src/containers/UserSettings/settings.tsx
```diff
@@ -13,6 +13,7 @@ import {
     THEME_KEY,
     USE_CLUSTER_BALANCER_AS_BACKEND_KEY,
     USE_DIRECTORY_OPERATIONS,
+    USE_HIDE_DOMAIN_DATABASE_KEY,
```
artemmufazalov, line 16: Why do you add 'use' everywhere? In other cases (`USE_CLUSTER_BALANCER_AS_BACKEND_KEY`, `USE_DIRECTORY_OPERATIONS`) it's a verb, here you already have a verb 'hide'

src/containers/UserSettings/i18n/en.json
astandrik: discussed with Alexey, he approved your suggestion

src/containers/UserSettings/settings.tsx
astandrik: moved to general

src/containers/UserSettings/settings.tsx
astandrik: now key is SHOW_DOMAIN_DATABASE

artemmufazalov (approved):

Issue #1077, opened 2024-07-29
Fetch available features info/versions from server


antonkovalenko:
UI must fetch `/capabilities` in the beginning of the session and store it for future checks.
This response will be used for checking whether server has particular feature or not?

antonkovalenko:
This issue is required for Stop Running Query ticket

PR #1097, opened 2024-07-31
feat: use features versions from backend

artemmufazalov:
Closes: #1077

To illustrate new handler, replaced two settings with feature versions from backend

Stand: https://nda.ya.ru/t/B1vOjA4N77KBQK


src/store/reducers/capabilities/capabilities.ts
```diff
@@ -0,0 +1,33 @@
+import {createSelector} from '@reduxjs/toolkit';
+
+import type {Capability} from '../../../types/api/capabilities';
+import type {RootState} from '../../defaultStore';
+
+import {api} from './../api';
+
+export const capabilitiesApi = api.injectEndpoints({
+    endpoints: (build) => ({
+        getClusterCapabilitie: build.query({
```
Raubzeug, line 10: typo getClusterCapabilitie`s`

src/store/reducers/capabilities/capabilities.ts
```diff
@@ -0,0 +1,33 @@
...
+            queryFn: async () => {
+                try {
+                    const data = await window.api.getClusterCapabilities();
+                    const {Capabilities} = data;
+                    return {data: Capabilities};
```
Raubzeug, line 15: if we don't need any checks here, we may  just return `{data}`

Raubzeug (changes requested):

Raubzeug (approved):

PR #1127, opened 2024-08-07
feat: search in queries history & saved

sareyu:


sareyu:
image

Search is performed by query text field, case sensitive

src/containers/Tenant/Query/QueriesHistory/QueriesHistory.tsx
```diff
@@ -23,8 +27,9 @@ interface QueriesHistoryProps {
 
 function QueriesHistory({changeUserInput}: QueriesHistoryProps) {
     const dispatch = useTypedDispatch();
+    const [filter, setFilter] = React.useState('');
 
-    const queriesHistory = useTypedSelector(selectQueriesHistory);
+    const queriesHistory = useTypedSelector(selectQueriesHistory(filter));
```
Raubzeug, line 32: What do you think if we operate with filter in store? This allows to preserve filter value when tabs are switching, and all staff with filtering would be done inside redux selectors (it would be no need to pass `filter` there).

Raubzeug (changes requested):

Raubzeug:
It seems that search should be case insensitive, cause sql keywords are both valid - uppercase and lowercase.

src/containers/Tenant/Query/QueriesHistory/QueriesHistory.tsx
sareyu: fixed

Raubzeug (approved):

PR #1152, opened 2024-08-13
feat: table preview doesnt update on click

astandrik:


src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
```diff
@@ -327,6 +328,7 @@ export function ObjectSummary({
     };
 
     const onOpenPreview = () => {
+        dispatch(api.util.invalidateTags(['PreviewData']));
```
Raubzeug, line 331: I think we shouldn't cache this request at all. Every time we click on preview, data should be requested.

Raubzeug (changes requested):

src/containers/Tenant/ObjectSummary/ObjectSummary.tsx
astandrik: Any ideas of other ways how it can be achieved despite tag invalidation?

Raubzeug (dismissed):

astandrik:
@Raubzeug splitted to getSchemaControls and getSummaryControls

PR #1165, opened 2024-08-15
feat(diagnostics): render bloom filter better

### LINKED ISSUE #1080: Table Info tab – display BloomFilter size
When bloom filter is enabled. let's display its size close to the info that it's enabled.

And in TableStats section

Describe is supposed to be enriched in [#7189](https://github.com/ydb-platform/ydb/issues/7189)

kunga:
Resolves #1080

Also removed bloom filter enable from default create table

### DIFF
--- a/src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.tsx
+++ b/src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.tsx
@@ -1,6 +1,7 @@
 import {Text} from '@gravity-ui/uikit';
 import omit from 'lodash/omit';
 
+import {toFormattedSize} from '../../../../../components/FormattedBytes/utils';
 import type {InfoViewerItem} from '../../../../../components/InfoViewer';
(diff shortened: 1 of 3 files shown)


src/containers/Tenant/Diagnostics/Overview/TableInfo/prepareTableInfo.tsx
```diff
@@ -126,10 +128,12 @@ const prepareTableGeneralInfo = (PartitionConfig: TPartitionConfig, TTLSettings?
         }
     }
 
-    generalTableInfo.push({
-        label: i18n('label.bloom-filter'),
-        value: EnableFilterByKey ? i18n('enabled') : i18n('disabled'),
-    });
+    if (EnableFilterByKey != undefined) {
```
artemmufazalov, line 131: It's better to use helper `valueIsDefined` or at least compare strictly `!==`. In this case it seems that there is no difference, but it generally good practice to avoid possible errors further

artemmufazalov (changes requested):

artemmufazalov (approved):

PR #1168, opened 2024-08-15
feat(PDiskPage): add decommission button

### LINKED ISSUE #1092: pdisk page – add decommission control
### Design drafts

Could be found at YCDESIGN-7572

### What to do

On a PDISK page decommission status must be displayed. 

Image

As a user I want to click on decommision status to change it
if pdisk is any other state than NONE, status must be dispayed near PDISKID

Image

When I change decommission status confirmation request dialog must be displayed

Image

artemmufazalov:
Closes #1092 

Test stand: https://nda.ya.ru/t/FBNWq_Uf77bWwM

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1168/)

| 124 | 119 | 0 | 5 | 0 |

### Bundle Size: 🔺
Current: 78.91 MB | Main: 78.88 MB
Diff: +0.03 MB (0.04%)

<details>

</details>


src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+        if (onConfirmActionSuccess) {
+            setButtonLoading(true);
+
+            // Decommission needs some time to change
+            // Wait for some time to send request for updated data
+            await wait(5000);
+
+            try {
+                await onConfirmActionSuccess();
+            } catch {
```
astandrik, line 74: This silently swallows any errors. It should at least log the error or notify the user.

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    const handleConfirmActionSuccess = async () => {
+        setWithRetry(false);
+
+        if (onConfirmActionSuccess) {
+            setButtonLoading(true);
+
+            // Decommission needs some time to change
+            // Wait for some time to send request for updated data
+            await wait(5000);
```
astandrik, line 70: This assumes the decommission status will change within 5 seconds, which may not always be true. A better approach would be to poll the API until the status changes or a timeout is reached.

src/containers/PDiskPage/DecommisionLabel/DecommisionLabel.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface DecommisionLabelProps {
+    decommission?: EDecommitStatus;
+}
+
+function getDecommissionLabelText(decommission: string) {
+    return pDiskPageKeyset('decommission-label', {decommission});
+}
+
+export function DecommisionLabel({decommission}: DecommisionLabelProps) {
```
astandrik, line 14: double s

src/components/ButtonWithConfirmDialog/ButtonWithConfirmDialog.tsx
```diff
@@ -3,13 +3,15 @@ import React from 'react';
 import {Button, Popover} from '@gravity-ui/uikit';
```
astandrik, line 3: There don't appear to be any new tests added for the new components and functionality. Adding unit and integration tests would help catch potential issues early. 

Cant insist but would be great

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+    buttonDisabled,
+    popoverDisabled,
+}: DecommissionButtonProps) {
+    const [newDecommission, setNewDecommission] = React.useState<EDecommitStatus | undefined>();
+    const [buttonLoading, setButtonLoading] = React.useState(false);
+    const [withRetry, setWithRetry] = React.useState(false);
+
+    const handleConfirmAction = async (isRetry?: boolean) => {
+        setButtonLoading(true);
+        await onConfirmAction(newDecommission, isRetry);
```
astandrik, line 58: This doesn't handle errors properly. If onConfirmAction throws an error, setButtonLoading(false) will never be called, potentially leaving the button in a perpetual loading state. Wrap this in a try/catch and ensure setButtonLoading(false) is always called.

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+        },
+        {
+            text: pDiskPageKeyset('decommission-imminent'),
+            theme: 'danger',
+            action: () => setNewDecommission('DECOMMIT_IMMINENT'),
+            hidden: decommission === 'DECOMMIT_IMMINENT',
+        },
+    ];
+
+    const renderSwitcher: DropdownMenuProps<unknown>['renderSwitcher'] = (props) => {
```
astandrik, line 113: can we move it to distinct component?

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+    };
+
+    const handleConfirmActionError = (error: unknown) => {
+        setWithRetry(isErrorWithRetry(error));
+        setButtonLoading(false);
+    };
+
+    const items: DropdownMenuItem[] = [
```
astandrik, line 86: Can we move it outside component?

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
```diff
@@ -0,0 +1,157 @@
...
+            <CriticalActionDialog
+                visible={Boolean(newDecommission)}
+                header={pDiskPageKeyset('decommission-dialog-title')}
+                text={getDecommissionWarningText(newDecommission)}
+                withRetry={withRetry}
+                retryButtonText={pDiskPageKeyset('decommission-dialog-force-change')}
+                onConfirm={handleConfirmAction}
+                onConfirmActionSuccess={handleConfirmActionSuccess}
+                onConfirmActionError={handleConfirmActionError}
+                onClose={() => {
```
astandrik, line 151: better to useCallback for passing functions as props

src/containers/PDiskPage/PDiskPage.tsx
```diff
@@ -96,7 +101,33 @@ export function PDiskPage() {
         return undefined;
     };
 
-    const handleAfterRestart = () => {
+    const handleDecommissionChange = async (
+        newDecomissionStatus?: EDecommitStatus,
+        isRetry?: boolean,
+    ) => {
+        if (pDiskParamsDefined) {
+            return window.api
```
astandrik, line 109: its async function
therefore we can use async/await syntax

astandrik (changes requested):

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
artemmufazalov: It's wrapped inside `CriticalActionDialog`, `handleConfirmActionSuccess` and `handleConfirmActionError` are called on success and error respectively. But you are right, there is no need to `setButtonLoading(false)` here, since it's done in other handlers

src/components/ButtonWithConfirmDialog/ButtonWithConfirmDialog.tsx
artemmufazalov: Probably, I'll add tests later, but not in this PR - I want to finish functionality first

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
artemmufazalov: It supposed, that data refetch errors will be handled in the component, where data is used. Example:

Screenshot 2024-09-03 at 16 56 55

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
artemmufazalov: I don't like the idea to poll the handler, that collects entity data (`/pdisk/info` with data both from BSC and Node Whiteboard is the only way we can get decommit status now). 

So this approach is not the best, but works in most of the cases - in other cases you could get actual status after some more time with refresh button

src/containers/PDiskPage/DecommissionButton/DecommissionButton.tsx
astandrik: I suppose the answer is "No" =)

src/containers/PDiskPage/PDiskPage.tsx
```diff
@@ -96,7 +101,33 @@ export function PDiskPage() {
         return undefined;
     };
 
-    const handleAfterRestart = () => {
+    const handleDecommissionChange = async (
```
astandrik, line 104:
```
const handleDecommissionChange = async (
    newDecommissionStatus?: EDecommitStatus,
    isRetry?: boolean
): Promise<void | undefined> => {
    if (!pDiskParamsDefined) {
        return undefined;
    }

    try {
        const response = await window.api.changePDiskStatus({
            nodeId,
            pDiskId,
            force: isRetry,
            decommissionStatus: newDecommissionStatus,
        });

        if (response?.result === false) {
            const err = {
                statusText: response.error,
                retryPossible: response.forceRetryPossible,
            };
            throw err;
        }
    } catch (error) {
        // Re-throw the error to be handled by the caller
        throw error;
    }
};
```

astandrik (approved):

src/containers/PDiskPage/PDiskPage.tsx
artemmufazalov: Thanks, I didn't get your point. Your version looks more aligned with our codebase

astandrik (approved):

astandrik (approved):

PR #1169, opened 2024-08-15
feat: duration & endtime in queries history

sareyu:
image

Added duration and endime IF Statistics collection mode is set to full in query settings.
Otherwise I have to overwrite this setting, thats seems to be a bad idea

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1169/)

| 114 | 111 | 0 | 3 | 0 |

### Bundle Size: 🔺
Current: 79.00 MB | Main: 79.00 MB
Diff: +0.01 MB (0.01%)

<details>

</details>


src/containers/Tenant/Query/QueryEditor/QueryEditor.tsx
```diff
@@ -170,6 +172,10 @@ function QueryEditor(props: QueryEditorProps) {
             schema,
             enableTracingLevel,
             queryId,
+        }).then((res) => {
+            if ('data' in res) {
```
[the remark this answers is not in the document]
ValeraS: There is a `dispatch` in the second param of the `queryFn`.

artemmufazalov (approved):

src/containers/Tenant/Query/QueriesHistory/QueriesHistory.tsx
```diff
@@ -57,6 +59,23 @@ function QueriesHistory({changeUserInput}: QueriesHistoryProps) {
             sortable: false,
             width: 600,
         },
+        {
+            name: 'EndTime',
+            header: i18n('history.endTime'),
+            render: ({row}) =>
+                row.endTime ? formatDateTime(new Date(row.endTime as string).getTime()) : '-',
```
artemmufazalov, line 66: `row.endTime.toString()` instead of `row.endTime as string`, it more error proof

src/store/reducers/executeQuery.ts
```diff
@@ -201,6 +240,19 @@ export const executeQueryApi = api.injectEndpoints({
...
+
+                    const queryStats: QueryStats = {};
+                    if (data.stats) {
+                        const {DurationUs, Executions: [{FinishTimeMs}] = [{}]} = data.stats;
+                        queryStats.duration = DurationUs;
+                        queryStats.endTime = FinishTimeMs;
+                    } else {
+                        const now = Date.now();
+                        queryStats.duration = (now - timeStart) * 1000;
+                        queryStats.endTime = now;
```
artemmufazalov, line 252: What do you think about saving duration as `ms`? `us` is not very common thing, it's become important not to forget all that conversions (`*1000` or `parseUsToMs`).

In both cases I suggest adding metric to field name: `durationMs` or `durationUs`

artemmufazalov (changes requested):

src/store/reducers/executeQuery.ts
sareyu: I keep existing format from api. we could replace `us` to `ms` in all places, but usually I prefer to keep original data as is (if api returns us - it means there are cases where this precision is important)

artemmufazalov (approved):

PR #1307, opened 2024-09-16
feat: add PDisk error reason in VDisk page and popup

SammyVimes:
https://github.com/ydb-platform/ydb/pull/9302


artemmufazalov:
We added column "VDisks with PDisks" to our storage table, that can be turned on by table columns settings. So now it's possible to look at PDisk together with VDisk. This storage view is displayed on every storage entity page - VDisk, PDisk, StorageGroup pages.

Screenshot 2024-10-15 at 15 50 11

Additional PDisk fields in VDisk popups seems not needed anymore

PR #1321, opened 2024-09-19
feat(Storage): add columns setup

### LINKED ISSUE #1216: Make list of columns adjustable for a table on the storage tab
Add selector similar to Clusters table

Screenshot 2024-08-26 at 13 06 27

two steps:

* [ ] add column selector to the table
* [ ] #1249

artemmufazalov:
Closes #1216


src/containers/UserSettings/settings.tsx
```diff
@@ -90,11 +89,6 @@ export const invertedDisksSetting: SettingProps = {
     settingKey: INVERTED_DISKS_KEY,
     title: i18n('settings.invertedDisks.title'),
 };
-export const useAdvancedStorageSetting: SettingProps = {
```
artemmufazalov, line 93: Now user will be able to select desired columns view via columns setup

Raubzeug:
@artemmufazalov what do you think if we move column selector in Vdisk and Pdisk pages on a previous line like:

artemmufazalov:
> @artemmufazalov what do you think if we move column selector in Vdisk and Pdisk pages on a previous line like

I don't like the control on the tabs level, since it is connected to content of one tab and isn't connected to another - `PDiskPage` , or control changes depending on tab - `StorageGroupPage`

As for `VDiskPage` - I don't see much difference, I prefer to leave it the same way as for `PDiskPage` and `StorageGroupPage`

astandrik (approved):

Raubzeug (approved):

PR #1364, opened 2024-09-26
feat(PaginatedStorage): add grouping

artemmufazalov:
Closes: #1365

Stand: https://nda.ya.ru/t/_9dz2Iw678WGs4

To test grouping, enable `Use paginated tables` setting in experiments

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1364/)

| 124 | 122 | 0 | 2 | 0 |

### Bundle Size: 🔺
Current: 79.13 MB | Main: 79.08 MB
Diff: +0.04 MB (0.06%)

<details>

</details>

### DIFF
--- a/src/components/PaginatedTable/TableChunk.tsx
+++ b/src/components/PaginatedTable/TableChunk.tsx
@@ -17,6 +17,7 @@ const typedMemo: <T>(Component: T) => T = React.memo;
 interface TableChunkProps<T, F> {
     id: number;
     limit: number;
+    totalLength: number;
     rowHeight: number;
     columns: Column<T>[];
     filters?: F;
@@ -35,6 +36,7 @@ interface TableChunkProps<T, F> {
 export const TableChunk = typedMemo(function TableChunk<T, F>({
     id,
     limit,
+    totalLength,
     rowHeight,
     columns,
     fetchData,
@@ -101,7 +103,11 @@ export const TableChunk = typedMemo(function TableChunk<T, F>({
         }
     }, [currentData, isActive, onDataFetched]);
 
-    const dataLength = currentData?.data?.length || limit;
+    const chunkOffset = id * limit;
+    const remainingLenght = totalLength - chunkOffset;
+    const calculatedChunkLength = remainingLenght < limit ? remainingLenght : limit;
+
+    const dataLength = currentData?.data?.length || calculatedChunkLength;
 
     const renderContent = () => {
         if (!isActive) {
--- a/src/containers/Storage/Storage.tsx
+++ b/src/containers/Storage/Storage.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
 
-import {StringParam, useQueryParams} from 'use-query-params';
-
 import {AccessDenied} from '../../components/Errors/403';
 import {isAccessError} from '../../components/Errors/PageError/PageError';
 import {ResponseError} from '../../components/Errors/ResponseError';
@@ -12,26 +10,20 @@ import {
 } from '../../store/reducers/capabilities/hooks';
 import {useClusterBaseInfo} from '../../store/reducers/cluster/cluster';
 import type {NodesSortParams} from '../../store/reducers/nodes/types';
-import {VISIBLE_ENTITIES} from '../../store/reducers/storage/constants';
 import {filterGroups, filterNodes} from '../../store/reducers/storage/selectors';
 import {storageApi} from '../../store/reducers/storage/storage';
-import {storageTypeSchema, visibleEntitiesSchema} from '../../store/reducers/storage/types';
-import type {
-    StorageSortParams,
-    StorageType,
-    VisibleEntities,
-} from '../../store/reducers/storage/types';
+import type {StorageSortParams} from '../../store/reducers/storage/types';
 import {valueIsDefined} from '../../utils';
 import {useAutoRefreshInterval, useTableSort} from '../../utils/hooks';
-import {NodesUptimeFilterValues, nodesUptimeFilterValuesSchema} from '../../utils/nodes';
 import {useAdditionalNodeProps} from '../AppWithClusters/useClusterData';
 
-import {StorageControls} from './StorageControls/StorageControls';
+import {StorageGroupsControls, StorageNodesControls} from './StorageControls/StorageControls';
 import {StorageGroupsTable} from './StorageGroups/StorageGroupsTable';
 import {useStorageGroupsSelectedColumns} from './StorageGroups/columns/hooks';
 import {StorageNodesTable} from './StorageNodes/StorageNodesTable';
 import {useStorageNodesSelectedColumns} from './StorageNodes/columns/hooks';
 import {b} from './shared';
+import {useStorageQueryParams} from './useStorageQueryParams';
 import {defaultSortNode, getDefaultSortGroup} from './utils';
 
 import './Storage.scss';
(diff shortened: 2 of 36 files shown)


src/components/PaginatedTable/TableChunk.tsx
```diff
@@ -101,7 +103,11 @@ export const TableChunk = typedMemo(function TableChunk<T, F>({
         }
     }, [currentData, isActive, onDataFetched]);
 
-    const dataLength = currentData?.data?.length || limit;
+    const chunkOffset = id * limit;
+    const remainingLenght = totalLength - chunkOffset;
+    const calculatedChunkLength = remainingLenght < limit ? remainingLenght : limit;
```
artemmufazalov, line 108: This allows not to show 50 (limit) loading rows for the last chunk, when we know how many entities are left. Also it allows to set some predefined initial length to table

src/containers/Storage/PDisk/PDisk.scss
```diff
@@ -21,7 +21,7 @@
     }
 
     &__vdisks-item {
-        flex-basis: 5px;
+        flex-basis: 3px;
```
artemmufazalov, line 24: Allow `PDisk` to have more compact `VDisks` on it. It's needed to make `StorageNodes` to have the same row height, allowing switching from `variable` RDT rendering to `uniform`

src/containers/Storage/StorageNodes/columns/StorageNodesColumns.scss
```diff
@@ -9,6 +9,7 @@
         align-items: flex-end;
 
         width: max-content;
+        height: 40px;
```
artemmufazalov, line 12: Make PDisk row to have the same height. It's needed to make `StorageNodes` to have the same row height, allowing switching from `variable` RDT rendering to `uniform`

src/containers/Storage/Storage.tsx
```diff
@@ -230,36 +170,29 @@ export const Storage = ({database, nodeId, groupId, pDiskId}: StorageProps) => {
...
+                {isNodes ? (
+                    <StorageNodesControls
+                        withTypeSelector
+                        entitiesCountCurrent={storageNodes.length}
+                        entitiesCountTotal={nodesTotalCount}
+                        entitiesLoading={isLoading}
+                        columnsToSelect={storageNodesColumnsToSelect}
+                        handleSelectedColumnsUpdate={setStorageNodesSelectedColumns}
+                    />
+                ) : null}
```
artemmufazalov, line 194: Added different controls to the same component. As the next step `Storage` could be split into two components like `PaginatedStorage`

src/containers/Storage/UsageFilter/UsageFilter.tsx
```diff
@@ -1,100 +0,0 @@
...
-    className?: string;
-    value?: string[];
-    groups?: UsageFilterItem[];
-    onChange?: (value: string[]) => void;
-    debounce?: number;
-}
-
-const b = cn('usage-filter');
-
-export const UsageFilter = (props: UsageFilterProps) => {
```
artemmufazalov, line 29: `UsageFilter` has been disabled in our UI for months. The problem it solved now is solved by grouping, it can be deleted

Raubzeug:
Let's add padding on the right of the groups container.
Screenshot 2024-09-30 at 09 49 45

Raubzeug:
If filter is set, I would probably expand groups, that contain filtered entities.
Screenshot 2024-09-30 at 09 55 34

Raubzeug:
I think we should preserve selected GroupBy setting when switch between Groups and Nodes storage.

src/containers/Storage/TableGroup/TableGroup.tsx
```diff
@@ -0,0 +1,65 @@
...
+        setIsExpanded(expanded);
+    }, [expanded]);
+
+    const toggleCollapsed = () => {
+        setIsExpanded((value) => !value);
+    };
+
+    const renderTitle = () => {
+        return (
+            <div onClick={toggleCollapsed} className={b('title-wrapper')}>
```
Raubzeug, line 39: Lets try to use button here, otherwise the element isn't reachable from keyboard

Raubzeug (changes requested):

artemmufazalov:
> If filter is set, I would probably expand groups, that contain filtered entities

Groups are filtered as well, so all groups on the screen will have only filtered entities

artemmufazalov:
> > If filter is set, I would probably expand groups, that contain filtered entities
> 
> Groups are filtered as well, so all groups on the screen will have only filtered entities

There is an issue with groups being closed every time search changes, I will fix it

artemmufazalov:
> I think we should preserve selected GroupBy setting when switch between Groups and Nodes storage.

It's preserved if it's the property, that exist for both nodes and groups. You may try "Disk space usage".

However, if it's a property, that doesn't valid for one type of storage - it isn't used, so if groups will be grouped by `PoolName` and then storage type will change, storage nodes won't be grouped by `PoolName` - currently it is not possible

In the code single search query `groupBy` param is used for both types of storage:
```js
// StorageGroups
export const storageGroupsGroupByParamSchema = z
    .custom<
        GroupsGroupByField | undefined
    >((value) => STORAGE_GROUPS_GROUP_BY_PARAMS.includes(value))
    .catch(undefined);

// StorageNodes
export const storageNodesGroupByParamSchema = z
    .custom<NodesGroupByField | undefined>((value) => STORAGE_NODES_GROUP_BY_PARAMS.includes(value))
    .catch(undefined);

// useStorageQueryParams
const storageGroupsGroupByParam = storageGroupsGroupByParamSchema.parse(queryParams.groupBy);
const storageNodesGroupByParam = storageNodesGroupByParamSchema.parse(queryParams.groupBy);

```

src/containers/Storage/StorageControls/StorageControls.tsx
```diff
@@ -83,17 +83,19 @@ export function StorageGroupsControls({
...
-                    label={i18n('controls_group-by-placeholder')}
-                    width={240}
-                    defaultValue={
-                        storageGroupsGroupByParam ? [storageGroupsGroupByParam] : undefined
-                    }
-                    onUpdate={handleGroupBySelectUpdate}
-                    options={STORAGE_GROUPS_GROUP_BY_OPTIONS}
-                />
+                <React.Fragment>
+                    <Text variant="body-2">{i18n('controls_group-by-placeholder')}</Text>
```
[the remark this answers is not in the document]
artemmufazalov: Probably, it depends on screen size. Today I use my laptop and don't have this issue even on the current uikit version, but I had it on the bigger screen with different resolution

src/containers/Storage/TableGroup/TableGroup.tsx
artemmufazalov: Button applies it's own styles, that doesn't fit here, so I added `tabIndex` to `div`

Raubzeug:
> It's preserved if it's the property, that exist for both nodes and groups. You may try "Disk space usage".

I'm not sure this is a proper way to handle it... as for me, we have to different storage types. And we should treat them separately. All controls should belong to selected storage type.

src/containers/Storage/TableGroup/TableGroup.tsx
Raubzeug: For now I may focus, but it doesn't expand on Enter.
Lets use button css api, like in this [example](https://github.com/gravity-ui/uikit/blob/b46f53603a260c7699e5bcc5c6899ed8cf13776d/src/components/Button/__stories__/Button.stories.tsx#L203).

artemmufazalov:
> It seems you did a fix, for now expand state doesn't reset on search change. But what do you think if we automatically expand a group, if 1) search is not empty 2) only one group is in result.

I don't like the idea to expand something automatically.

This groups can be used not only to group entities, but also to view overall cluster statistics. By enabling some group by option, I can see, how cluster feels in term of that option.

Automatic expand does't really suit this case. There is no indication on how much table groups are rendered. We have only one group -> it's automatically expanded -> it's not obvious for me, that we have only one group until I scroll to bottom.

Also it's optimization -  content is not loaded and rendered until group is expanded.

To sum up, I don't think we should auto expand, but we can discuss it and implement in next iteration if needed

Raubzeug (approved):

PR #1426, opened 2024-10-08
fix(Authentication): handle login error properly

### LINKED ISSUE #1423: result of login operation ignored
if you login with any invalid user/password, login handler will return error, but UI ignores it.

artemmufazalov:
Closes #1423 

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1426/)

| 128 | 127 | 0 | 1 | 0 |

### Bundle Size: ✅
Current: 79.13 MB | Main: 79.13 MB
Diff: +0.08 KB (0.00%)

✅ Bundle size unchanged.

<details>

</details>

### DIFF
--- a/src/containers/Authentication/Authentication.tsx
+++ b/src/containers/Authentication/Authentication.tsx
@@ -24,7 +24,7 @@ function Authentication({closable = false}: AuthenticationProps) {
     const history = useHistory();
     const location = useLocation();
 
-    const [authenticate, {error, isLoading}] = authenticationApi.useAuthenticateMutation(undefined);
+    const [authenticate, {isLoading}] = authenticationApi.useAuthenticateMutation(undefined);
 
     const {returnUrl} = parseQuery(location);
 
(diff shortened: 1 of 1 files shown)


src/containers/Authentication/Authentication.tsx
```diff
@@ -24,7 +24,7 @@ function Authentication({closable = false}: AuthenticationProps) {
     const history = useHistory();
     const location = useLocation();
 
-    const [authenticate, {error, isLoading}] = authenticationApi.useAuthenticateMutation(undefined);
+    const [authenticate, {isLoading}] = authenticationApi.useAuthenticateMutation(undefined);
```
astandrik, line 27: Why didn't it work as expected?
I mean isn't auth mutation supposed to return error on error?

artemmufazalov: Mutation result is always `{data, error}`, it cannot be processed by catch without `.unwrap()`. In previous version code in `.then` always ran, that caused the bug with error being ignored. I could use only `.then` as well - `.then(({data, error}) -> {})`, but  form with explicit `.catch` looks better for me

astandrik (approved):

PR #1436, opened 2024-10-10
feat: improve appearing controls styles

### LINKED ISSUE #1045: Clipboard button
For now clipboard button, that appears on hover, takes reserved place even if it is invisible.
Lets improve layout and place additional controls (clipboard button, outer link, etc.) over the content (so, the content should not be shrinked by appeared controls)

Raubzeug:
closes #1045 

[Stand](https://nda.ya.ru/t/BujeumfW78xr8D)

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1436/)

| 134 | 134 | 0 | 0 | 0 |

### Bundle Size: 🔽
Current: 78.94 MB | Main: 78.97 MB
Diff: 0.03 MB (-0.04%)

✅ Bundle size decreased.

<details>

</details>


src/components/DeveloperUILinkButton/DeveloperUILinkButton.tsx
```diff
@@ -7,16 +8,30 @@ import './DeveloperUILinkButton.scss';
...
+    href,
+    visible = false,
+    className,
+    size = 's',
+}: DeveloperUiLinkProps) {
     return (
-        <Button size="s" href={href} className={b({visible}, className)} target="_blank">
-            <Icon data={ArrowUpRightFromSquare} />
+        <Button size={size} href={href} className={b({visible}, className)} target="_blank">
+            <Icon data={ArrowUpRightFromSquare} size={buttonSizeToIconSize[size]} />
```
astandrik, line 34: I suppose we can use Button Icons as here

```
       <Button
                view="flat"
                onClick={() => {
                    dispatch(api.util.invalidateTags(['All']));
                }}
                extraProps={{'aria-label': i18n('Refresh')}}
            >
                <Button.Icon>
                    <ArrowsRotateLeft />
                </Button.Icon>
            </Button>
 ```
 
 
 and icon size should be set automatically according to button size

Comment (Raubzeug) [reply src/components/DeveloperUILinkButton/DeveloperUILinkButton.tsx:34]: Discussed with uikit team, button doesn't affect it's icon size. We should explicitly set icon size we want.

File: src/containers/App/App.scss [hunk 2]
```diff
@@ -93,9 +93,14 @@ body,
 .ydb-paginated-table__row,
 .ydb-tree-view__item {
     &:hover {
```
artemmufazalov, line 95: Since you add padding only here, you should also add padding for focused state. Now if I select button with Tab, it won't have right padding

src/components/EntityStatus/EntityStatus.tsx
```diff
@@ -95,18 +94,21 @@ export function EntityStatus({
...
+                        size="xs"
+                        view="normal"
+                        className={b('clipboard-button', {
+                            visible: clipboardButtonAlwaysVisible,
+                        })}
+                    />
+                )}
+                {additionalControls && (
+                    <span className={b('additional-controls')}>{additionalControls}</span>
+                )}
```
artemmufazalov, line 110: `DeveloperUIButton` is controlled by `display: none` and `display: inline-block`. It cannot be selected with Tab. Two ways possible:

1. All buttons in the UI should be keyboard accessible, so this `DeveloperUIButton` button should be updated too

2. We don't need to make buttons in table keyboard accessible (for me in tables it doesn't make much sense anyway), so we can stick to previous solution, but update only `ClipboardButton`

src/containers/Tablets/Tablets.tsx
```diff
@@ -71,6 +71,7 @@ function getColumns({database}: {database?: string}) {
                         additionalControls={
                             <DeveloperUILinkButton
                                 href={createTabletDeveloperUIHref(row.TabletId)}
+                                size="xs"
```
artemmufazalov, line 74: You should decrease `TabletId` column width, it's contains unneeded empty space now (I guess 190 will be just fine)

artemmufazalov (changes requested):

src/containers/App/App.scss
Raubzeug: I refactored styles. Wrapper is only needed on hover. If buttons are focused from keyboard, they should be shown separately.

src/components/EntityStatus/EntityStatus.tsx
Raubzeug: Lets try the hardest way number 1. I hope I've found a solution.

artemmufazalov (approved):

artemmufazalov (approved):

Issue #1190, opened 2024-08-20
implement simple and narrow vertical progress bar


adameat:
to display usages in a compact way
![image](https://github.com/user-attachments/assets/00907184-4e21-4799-81f0-3e0f91ad0212)

use it to display total memory and total cpu on node table view

PR #1560, opened 2024-10-30
feat: implement simple and narrow vertical progress bar

astandrik:
Stand [Stand](http://ydb-vla-dev02-002.search.yandex.net:8765/monitoring/cluster/nodes)

Closes https://github.com/ydb-platform/ydb-embedded-ui/issues/1190

📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1560/)

| 134 | 134 | 0 | 0 | 0 |

### Bundle Size: 🔺
Current: 79.20 MB | Main: 79.19 MB
Diff: +0.01 MB (0.01%)

<details>

</details>


src/components/nodesColumns/columns.tsx
```diff
@@ -143,6 +201,50 @@ export function getCpuColumn<T extends {PoolStats?: TPoolStats[]}>(): Column<T>
...
+                        colorizeProgress
+                        percents
+                        vertical
+                    />
+                </CellWithPopover>
+            );
+        },
+        align: DataTable.LEFT,
+        width: 85,
+        resizeMinWidth: 40,
```
Raubzeug, line 245: It looks strange when resized. Let's do either with fixed bar width or non-resizable column.
Screenshot 2024-10-31 at 18 16 58

Raubzeug (changes requested):

src/components/nodesColumns/columns.tsx
```diff
@@ -111,6 +118,57 @@ export function getMemoryColumn<
         resizeMinWidth: 170,
     };
 }
+
+export function getRAMColumn<T extends {MemoryUsed?: string; MemoryLimit?: string}>(): Column<T> {
```
artemmufazalov, line 122: There is no sorting for new columns if table is paginated.

Currently, for every column `name` is checked with `isSortableNodesProperty`, if `name` is in the list, the column is sortable and it's name is sent to backend as sort param.

I'm not sure what's to do in this case, since there are two columns with different ids, but they have the same sort fields (`CPU` and `Memory`). 

Probably, you can use not column `name`, but add separate `sortField` property to every column object (in such case it's better to do it in a separate PR, since it's a lot of refactoring). Maybe there is a better and easier solution

astandrik: created dictionary for sort values

not sure if it's the best one in the world, but as far as we decided to rename CPU to Pools looks like there are not so many choices left...

src/components/nodesColumns/columns.tsx
astandrik: reworked

Raubzeug (approved):

PR #1661, opened 2024-11-20
refactor: merge query result components

artemmufazalov:


Raubzeug:
Screenshot 2024-11-22 at 09 40 04

It seems Stats are lost when I "Run" query.

artemmufazalov:
> It seems Stats are lost when I "Run" query. For now we show it even if it is `null`. Probably we shouldn't do it. But I also suggest to remove radio button if only result is shown.

What do you think, if we will show all tabs, like we do for explain, but if something is empty, we will show some stub message?

Screenshot 2024-11-22 at 12 01 33

Screenshot 2024-11-22 at 12 01 37

Screenshot 2024-11-22 at 12 01 41

Screenshot 2024-11-22 at 12 01 41

Raubzeug:
> What do you think, if we will show all tabs, like we do for explain, but if something is empty, we will show some stub message?

Deal!

Raubzeug (approved):

PR #1719, opened 2024-11-28
fix(Versions): tune default color

Raubzeug:


src/utils/versions/getVersionsColors.ts
```diff
@@ -12,6 +12,7 @@ export const hashCode = (s: string) => {
 // TODO: colors used in charts as well, need to move to constants
 // 11 distinct colors from https://mokole.com/palette.html
 export const COLORS = [
+    '#008000', // green
```
ValeraS, line 15: Before all the changes, there were 11 colors (see comment in the code above). Let's not change the number of colors.

ValeraS (changes requested):

src/utils/versions/getVersionsColors.ts
```diff
@@ -12,19 +12,19 @@ export const hashCode = (s: string) => {
...
     '#4169e1', // royalblue
     '#ffd700', // gold
-    '#ff8c00', // darkorange
```
ValeraS, line 17: Why is this deleted?

ValeraS (changes requested):

src/utils/versions/getVersionsColors.ts
Raubzeug: unintentionally :(

PR #1748, opened 2024-12-11
feat: change uptime format

### LINKED ISSUE #1067: change how uptime is displayed
currently we have:

image

I suggest to change how it displayed to simplify readibilty by making connection between value of uptime with length of the text representation:

everything below 1 minute we show just as seconds without leading zero:

  5s
10s
30s

after 1 minute we show short time representation:

  1:05
  5:10
15:30

after 1 hour we show longer time representation:

  1:30:20
  5:05:05
15:30:40

after 24 hours we adding number of the days
  
  1d 01:30:20
  1d 15:05:05
12d 05:05:05

of course everything with right-alignment. and we dont show leading zeros on the leftmost elements.

I expect it to be something like this:

image

artemmufazalov:
Closes #1067 

Screenshot 2024-12-11 at 15 19 58

Stand (there is hardcoded uptime in Nodes table for demonstration): https://nda.ya.ru/t/jnRsXjJ47AKdRa


src/utils/dataFormatters/dataFormatters.ts
```diff
@@ -40,19 +40,33 @@ export const stringifyVdiskId = (id?: TVDiskID | TVSlotId) => {
...
 };
 
-export const formatUptimeInSeconds = (seconds: number) => {
-    const days = Math.floor(seconds / DAY_IN_SECONDS);
-    const remain = seconds % DAY_IN_SECONDS;
+/**
+ * It works well only with positive values,
+ * if you want to get negative formatted uptime, use some wrapper like getDowntimeFromDateFormatted
+ */
+export function formatUptimeInSeconds(seconds: number) {
```
Raubzeug, line 47: lets handle somehow negative values. As for me this function should work properly with all values. And we should use it for negative and positive uptime.

Raubzeug (changes requested):

Raubzeug (approved):

Issue #730, opened 2024-02-22
VDisks in 2 rows


artemmufazalov:
On some clusters we have more than 18 VDisks on 1 PDisks. With this amount they are wrapped to a second row, but without preserving relative size (VDIsk width corresponds to its relative data size). So VDisks on the second row have incorrect relative size. 

Also this causes rows to have different height, that leads to problems with pagination / virtualization

Solution:

Never display vdisks in two rows, automatically extend pdisk size to hold all vdisks

![Screen Shot 2024-02-22 at 18 21 00](https://github.com/ydb-platform/ydb-embedded-ui/assets/67755036/f9a9f9e2-f059-418d-85ce-7c80bebb1b99)

artemmufazalov:
`VDisk` compact min size was decreased, it's needed to check whether this issue is still valid for our clusters

artemmufazalov:
https://github.com/ydb-platform/ydb/issues/10160

PR #1758, opened 2024-12-16
feat: vdisks in 2 rows

astandrik:
Closes https://github.com/ydb-platform/ydb-embedded-ui/issues/730

[Stand](https://nda.ya.ru/t/rvEWCRoK7ASsvB)

use parameters ( /cluster/storage?type=nodes&mocks=1&pdisks=10&vdisksPerPDisk=100 )

to play with markup

This PR is really only about changing two css files

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1758/)

  | 222 | 222 | 0 | 0 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: ✅
  Current: 65.88 MB | Main: 65.88 MB
  Diff: +2.68 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>


src/mocks/storage/nodes.ts
```diff
@@ -0,0 +1,158 @@
+import {EFlag} from '../../types/api/enums';
```
astandrik, line 1: for test only (will delete before merge)

src/containers/Storage/PDisk/PDisk.scss
```diff
@@ -1,28 +1,34 @@
...
+
+        white-space: nowrap;
     }
 
     &__vdisks-item {
-        flex-basis: 3px;
-        flex-shrink: 0;
+        flex: 0 0 3px;
+
+        min-width: 3px;
```
artemmufazalov, line 31: Your solution works well only in the case, where all pdisks have the same amount of vdisks, it's not always true.

Previously we assumed, that all pdisks should be of the same size, so `MaximumSlotsPerDisk` was introduced so it will be possible to calculate pdisk size for paginated table (something like `pDiskSize = MaximumSlotsPerDisk * VDISK_SIZE`) and use it for all pdisks in this table (even if the don't have any vdisks)

Screenshot 2024-12-17 at 13 46 51

artemmufazalov (changes requested):

src/containers/Storage/PDisk/PDisk.scss
astandrik: fixed

Also added random vdisks count generation to pdisks in mocks

src/store/reducers/storage/utils.ts
```diff
@@ -232,7 +258,8 @@ export const prepareStorageNodesResponse = (data: TNodesInfo): PreparedStorageRe
         return undefined;
     }).filter((group): group is TableGroup => Boolean(group));
 
-    const preparedNodes = Nodes?.map(prepareStorageNodeData);
+    const maximumSlots = calculateMaximumSlotsPerDisk(Nodes, MaximumSlotsPerDisk);
+    const preparedNodes = Nodes?.map((node) => prepareStorageNodeData(node, maximumSlots));
```
artemmufazalov, line 262: It looks like different table chunks could have different pdisk size with such calculations, since `prepareStorageNodesResponse` is called for every chunk

artemmufazalov (changes requested):

src/store/reducers/storage/utils.ts
artemmufazalov: It handles corner case of the scenario, which it a corner case by itself. Decided in pm to leave it as it is in PR

artemmufazalov (approved):

PR #1864, opened 2025-01-21
fix(UptimeViewer): do not show StartTime if DisconnectTime present

### LINKED ISSUE #1858: Do not show StartTime in Uptime popup if there is DisconnectTime
`StartTime` may be cached somewhere on backend and we could receive both `StartTime` and `DisconnectTime` for disconnected node, in this case we have both values in the tooltip with different timestamps

artemmufazalov:
Closes #1858


src/components/UptimeViewer/UptimeViewer.tsx
```diff
@@ -17,33 +19,32 @@ interface NodeUptimeProps {
...
-                            {formatDateTime(StartTime, {withTimeZone: true})}
-                        </DefinitionList.Item>
-                    ) : null}
-                    {DisconnectTime ? (
-                        <DefinitionList.Item key={'DisconnectTime'} name={i18n('disconnect-time')}>
-                            {formatDateTime(DisconnectTime, {withTimeZone: true})}
-                        </DefinitionList.Item>
-                    ) : null}
-                </DefinitionList>
-            }
```
artemmufazalov, line 46: Previously I assumed, that if node is disconnected, there will be no `StartTime` in the response. However, `StartTime` may be cached somewhere on backend and we could receive both `StartTime` and `DisconnectTime` for disconnected node, in this case we have both values in the tooltip with different timestamps. After this fix `StartTime` will be in popover only if there is no `DisconnectTime`

src/components/UptimeViewer/UptimeViewer.tsx
```diff
@@ -17,33 +19,32 @@ interface NodeUptimeProps {
...
+            content={<DefinitionList responsive>{content}</DefinitionList>}
```
Raubzeug, line 47: let's add a case if `content` is `null`

Raubzeug (changes requested):

Raubzeug (approved):

PR #1922, opened 2025-02-10
fix: code assistant option

astandrik:


Raubzeug (dismissed):

astandrik:
@Raubzeug used factory instance creation instead of hook for register/unregister
Checked on internal instance that it actually works

@artemmufazalov - you may find it interesting to take a glance

src/containers/Tenant/Query/QueryEditor/YqlEditor.tsx
```diff
@@ -71,20 +75,18 @@ export function YqlEditor({
...
-        config: {
-            language: YQL_LANGUAGE_ID,
-        },
-    });
+    const {monacoGhostConfig, prepareUserQueriesCache} = useCodeAssistHelpers();
+
+    React.useEffect(() => {
+        if (monacoGhostInstance && isCodeAssistEnabled) {
+            monacoGhostInstance.register(monacoGhostConfig);
+            prepareUserQueriesCache();
```
Raubzeug, line 83: Should we prepare cache every time any useEffect dependency changes?

src/containers/Tenant/Query/QueryEditor/YqlEditor.tsx
```diff
@@ -45,8 +46,11 @@ export function YqlEditor({
 }: YqlEditorProps) {
     const input = useTypedSelector(selectUserInput);
     const dispatch = useTypedDispatch();
+    const [monacoGhostInstance, setMonacoGhostInstance] =
```
Raubzeug, line 49: let's not use state for editor instance? Will `window.ydbEditor` be enough?

Raubzeug (approved):

PR #1930, opened 2025-02-11
fix(Storage): fix disks view for degraded group

### LINKED ISSUE #1929: Fix degraded group Disks display
1. Do not add space between disks if group has erasure `none` or `block-4-2` even if there is no info about disks DC
2. In Disks view display PDisk with no data for every VDisk even if there is no PDisk field in data
3. Make VDisks have average size if there is no data

Image

artemmufazalov:
Closes #1929 

Stand with the example of degraded group: https://nda.ya.ru/t/H7kJOQl57BtQVV

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1930/)

  | 262 | 259 | 0 | 3 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: ✅
  Current: 80.33 MB | Main: 80.32 MB
  Diff: +1.20 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/src/containers/Storage/Disks/Disks.tsx
+++ b/src/containers/Storage/Disks/Disks.tsx
@@ -3,9 +3,10 @@ import React from 'react';
 import {Flex, useLayoutContext} from '@gravity-ui/uikit';
 
 import {VDisk} from '../../../components/VDisk/VDisk';
-import {valueIsDefined} from '../../../utils';
+import type {Erasure} from '../../../types/api/storage';
 import {cn} from '../../../utils/cn';
 import type {PreparedVDisk} from '../../../utils/disks/types';
+import {isNumeric} from '../../../utils/utils';
 import {PDisk} from '../PDisk';
 import type {StorageViewContext} from '../types';
 import {isVdiskActive, useVDisksWithDCMargins} from '../utils';
@@ -19,12 +20,13 @@ const VDISKS_CONTAINER_WIDTH = 300;
 interface DisksProps {
     vDisks?: PreparedVDisk[];
     viewContext?: StorageViewContext;
+    erasure?: Erasure;
 }
 
-export function Disks({vDisks = [], viewContext}: DisksProps) {
+export function Disks({vDisks = [], viewContext, erasure}: DisksProps) {
     const [highlightedVDisk, setHighlightedVDisk] = React.useState<string | undefined>();
 
-    const vDisksWithDCMargins = useVDisksWithDCMargins(vDisks);
+    const vDisksWithDCMargins = useVDisksWithDCMargins(vDisks, erasure);
 
     const {
         theme: {spaceBaseSize},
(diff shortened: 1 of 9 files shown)


src/containers/Storage/Disks/Disks.tsx
```diff
@@ -89,7 +91,7 @@ function VDiskItem({
     const vDiskId = vDisk.StringifiedId;
 
     // show vdisks without AllocatedSize as having average width (#1433)
-    const minWidth = valueIsDefined(vDiskToShow.AllocatedSize) ? undefined : unavailableVDiskWidth;
+    const minWidth = isNumeric(vDiskToShow.AllocatedSize) ? undefined : unavailableVDiskWidth;
```
artemmufazalov, line 94: AllocatedSize could be numeric or `NaN`, but not `undefined`

src/containers/Storage/Disks/Disks.tsx
```diff
@@ -55,7 +57,7 @@ export function Disks({vDisks = [], viewContext}: DisksProps) {
             <div className={b('pdisks-wrapper')}>
                 {vDisks?.map((vDisk, index) => (
                     <PDiskItem
-                        key={vDisk?.PDisk?.StringifiedId}
+                        key={vDisk?.PDisk?.StringifiedId || index}
```
artemmufazalov, line 60: There could be no proper id, use index in such case

src/containers/Storage/utils/index.ts
```diff
@@ -84,12 +85,21 @@ export function getStorageGroupsInitialEntitiesCount(
     return DEFAULT_ENTITIES_COUNT;
 }
 
-export function useVDisksWithDCMargins(vDisks: PreparedVDisk[] = []) {
+function isErasureWithDifferentDC(erasure?: Erasure) {
+    return erasure === 'mirror-3-dc' || erasure === 'mirror-3of4';
```
astandrik, line 89: what does these magic constant strings mean?

comment would be nice (and may be move to actual constants)

src/types/api/storage.ts
```diff
@@ -228,6 +228,8 @@ export interface TStoragePDisk {
     Whiteboard?: TPDiskStateInfo;
 }
 
+export type Erasure = 'none' | 'block-4-2' | 'mirror-3-dc' | 'mirror-3of4';
```
astandrik, line 231: maybe enum and use it in isErasureWithDifferentDC ?

src/containers/Storage/Disks/Disks.tsx
astandrik: in which cases there may be no StringifiedId and why cant we always use index

I dont mean to rework anything - I just dont understand (:

src/types/api/storage.ts
artemmufazalov: I don't like enums, you have to import them everywhere to satisfy TS, while for union string types every option will be suggested by autocomplete

src/containers/Storage/Disks/Disks.tsx
artemmufazalov: 1. If there is no data about PDisk, then there would be no id, so we use index to prevent undefined ids - with undefined ids there is a bug where new disks added on every rerender
2. Generally it's not a good practice to use indexes as keys, since they could cause some unexpected behaviour if array is reordered or some elements added or deleted (and it's not recommended in react docs). So we use unique ids everywhere where it's possible. For the case of missing vdisks' pdisks there is no much difference, since they have no data and displayed the same way

astandrik (approved):

PR #1946, opened 2025-02-13
fix(ObjectSummary): refresh tabs with tree refresh, disable autorefresh

### LINKED ISSUE #1906: autorefresh on query tab keeps refreshing diagnostics (describe)
Image

artemmufazalov:
Closes #1906 

- Do not autorefresh Preview
- Do not autorefresh Overview, ACL and Schema
- Refresh Overview, ACL and Schema with tree refresh button

Overview and Schema are still affected by `Diagnostics` manual refresh, since the same apis are used in `Diagnostics`. It does not make much difference since we refresh `overviewApi` in `Diagnostics` anyway (it is used in `Tenant` component). The only unneeded refresh is `viewSchema` - if it is opened in left panel, it will be refreshed by `Diagnostics` manual refresh since this api used both in schema tabs in diagnostics. I did not find a simple solution that prevents it, so I would leave it as before, as for me this case is rare enough, while request is not heavy (`select * limit 0`)

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1946/)

  | 262 | 262 | 0 | 0 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: ✅
  Current: 80.32 MB | Main: 80.32 MB
  Diff: +0.12 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/src/containers/Tenant/Tenant.tsx
+++ b/src/containers/Tenant/Tenant.tsx
@@ -12,7 +12,7 @@ import {selectSchemaObjectData} from '../../store/reducers/schema/schema';
 import type {AdditionalNodesProps, AdditionalTenantsProps} from '../../types/additionalProps';
 import {cn} from '../../utils/cn';
 import {DEFAULT_IS_TENANT_SUMMARY_COLLAPSED, DEFAULT_SIZE_TENANT_KEY} from '../../utils/constants';
-import {useAutoRefreshInterval, useTypedDispatch, useTypedSelector} from '../../utils/hooks';
+import {useTypedDispatch, useTypedSelector} from '../../utils/hooks';
 import {isAccessError} from '../../utils/response';
 
 import ObjectGeneral from './ObjectGeneral/ObjectGeneral';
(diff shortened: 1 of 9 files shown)


src/containers/Tenant/Tenant.tsx
```diff
@@ -95,12 +94,7 @@ export function Tenant(props: TenantProps) {
         currentData: currentItem,
         error,
         isLoading,
-    } = overviewApi.useGetOverviewQuery(
-        {path, database: tenantName},
-        {
-            pollingInterval: autoRefreshInterval,
-        },
-    );
+    } = overviewApi.useGetOverviewQuery({path, database: tenantName});
```
artemmufazalov, line 97: We make this request to receive schema object types (they are not changed unless object is deleted and new object with the same name is created) and check access rights, so we don't need autorefresh here

src/store/reducers/overview/overview.ts
```diff
@@ -51,7 +51,7 @@ export const overviewApi = api.injectEndpoints({
                 }
             },
             keepUnusedDataFor: 0,
-            providesTags: ['All'],
+            providesTags: ['All', 'SchemaTree'],
```
artemmufazalov, line 54: Overview and Schema are still affected by Diagnostics manual refresh, since the same apis are used in Diagnostics. I did not find a simple solution that prevents it, so I would leave it as before, since it does not look as a big issue

astandrik (approved):

PR #1992, opened 2025-03-03
feat: add QUERY_TECHNICAL_MARK to all UI queries

### LINKED ISSUE #1490: Add `QUERY_TECHNICAL_MARK` to all UI queries
```
const QUERY_TECHNICAL_MARK = '/*UI-QUERY-EXCLUDE*/';
```

1. It will help to sort out such queries from our query tables
2. It will make easier to determine, whether it's UI request in case of some troubles

artemmufazalov:
Closes #1490 

Why to mark queries?
1. UI request to system tables won't be displayed in queries tables
2. It will be easier to distinguish UI queries from users' queries in case of incidents


src/store/reducers/executeTopQueries/executeTopQueries.ts
```diff
@@ -49,6 +48,25 @@ LIMIT 100
...
+
+    const orderBy = prepareOrderByFromTableSort(sortOrder);
+
+    return `${QUERY_TECHNICAL_MARK}
+SELECT
+    UserSID, 
+    QueryStartAt, 
+    Query as QueryText, 
+    ApplicationName
+FROM \`${path}/.sys/query_sessions\`
```
artemmufazalov, line 64: In case of nested DB previous query always returned running queries of domain database, so I added path

src/store/reducers/shardsWorkload/shardsWorkload.ts
```diff
@@ -63,27 +65,28 @@ function createShardQueryHistorical(
     PeakTime,
     InFlightTxCount,
     IntervalEnd
-FROM \`.sys/top_partitions_one_hour\`
+FROM \`${database}/.sys/top_partitions_one_hour\`
```
astandrik, line 68: If database is empty (as far as I can see - it can be '' because we are checking its value in ternary above) - will /.sys/... be okay as select source?

src/store/reducers/shardsWorkload/shardsWorkload.ts
```diff
@@ -128,12 +131,12 @@ export const shardApi = api.injectEndpoints({
                         {
                             query:
                                 filters?.mode === EShardsWorkloadMode.Immediate
-                                    ? createShardQueryImmediate(path, sortOrder, database)
+                                    ? createShardQueryImmediate(path, database, sortOrder)
                                     : createShardQueryHistorical(
                                           path,
+                                          database,
```
astandrik, line 137: As I see here path and database can exist simulteneously
Could you please explain the real difference (as somewhere we change path to database and somewhere not)

sareyu (commented):
LGTM

src/store/reducers/shardsWorkload/shardsWorkload.ts
artemmufazalov: Example: we look at shards of a table `db1/table1`
Database is `db1`, it should be used as a request param, system tables in `db1/.sys/`, but the result should be filtered, so only shards of current path `db1/table1` will be shown

artemmufazalov: `path` - path to schema object
`database` - database, where this schema object exist

src/store/reducers/shardsWorkload/shardsWorkload.ts
artemmufazalov: `.sys` will work, `/.sys` - won't. However, we alway sent request to specific database, otherwise queries may not work at all (currently I have errors when I sent requests without defined database).

I think it's better to remove ternaries - they have no sense.

Also I found more complex issue, we don't determine db correctly for nested entities, so for nested databases we have wrong db anyway. So actually, there is no difference, whether we pass database or not, `.sys` path could be relative when db is passed. Probably I revert the changes where I add database to `.sys` path

artemmufazalov: Removed unneeded ternaries, database should never be empty or request won't be successful anyway. BTW, these ternaries were not needed even in case of empty string database, YQL processed it correctly.

Used relative paths everywhere: both in query text and api request we use the same param, so it never differs. If we use relative path, the query will use path from the database, that we passed.

sareyu (approved):

PR #1997, opened 2025-03-06
fix: fix undefined process in package

artemmufazalov:
When ydb-embedded-ui is used as package, there is an error, that `process` is not defined. That's why I import package json directly

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/1997/)

  | 264 | 261 | 0 | 2 | 1 |

  
  <details>
  <summary>Test Changes Summary ⏭️1 </summary>

  #### ⏭️ Skipped Tests (1)
1. Streaming query shows some results and banner when stop button is clicked (tenant/queryEditor/queryEditor.test.ts)

  </details>

  ### Bundle Size: ✅
  Current: 80.81 MB | Main: 80.81 MB
  Diff: 0.15 KB (-0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/config-overrides.js
+++ b/config-overrides.js
@@ -2,15 +2,12 @@ const path = require('path');
 
 const MiniCSSExtractPlugin = require('mini-css-extract-plugin');
 const MonacoWebpackPlugin = require('monaco-editor-webpack-plugin');
-const webpack = require('webpack');
 const srcRoot = path.resolve(__dirname, 'src');
 const uiKitRoot = path.resolve(__dirname, 'node_modules/@gravity-ui/uikit');
 const antlr4C3Root = path.resolve(__dirname, 'node_modules/antlr4-c3');
 const websqlRoot = path.resolve(__dirname, 'node_modules/@gravity-ui/websql-autocomplete');
 const antlr4ngRoot = path.resolve(__dirname, 'node_modules/antlr4ng');
 const uiKitIconsRoot = path.resolve(__dirname, 'node_modules/@gravity-ui/icons');
-
-const packageJson = require('./package.json');
 
 module.exports = {
     webpack: (config, env) => {
@@ -45,13 +42,6 @@ module.exports = {
                         entry: 'monaco-yql-languages/build/monaco.contribution',
                     },
                 ],
-            }),
-        );
-
-        // Add DefinePlugin to expose just the version
-        config.plugins.push(
-            new webpack.DefinePlugin({
-                'process.env.UI_VERSION': JSON.stringify(packageJson.version),
             }),
         );
 
--- a/package.json
+++ b/package.json
@@ -164,8 +164,7 @@
     "source-map-explorer": "^2.5.3",
     "stylelint": "^15.11.0",
     "ts-jest": "^29.2.5",
-    "typescript": "^5.7.3",
-    "webpack": "^5.98.0"
+    "typescript": "^5.7.3"
   },
   "peerDependencies": {
     "monaco-yql-languages": ">=1.3.0",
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -164,8 +164,7 @@
     "source-map-explorer": "^2.5.3",
     "stylelint": "^15.11.0",
     "ts-jest": "^29.2.5",
-    "typescript": "^5.7.3",
-    "webpack": "^5.98.0"
```
artemmufazalov, line 168: There is no need to install it explicitly

src/containers/UserSettings/settings.tsx
```diff
@@ -141,7 +143,7 @@ export const autocompleteOnEnterSetting: SettingProps = {
 export const interfaceVersionInfoField: SettingsInfoFieldProps = {
     title: i18n('settings.about.interfaceVersionInfoField.title'),
     type: 'info',
-    content: process.env.UI_VERSION,
+    content: packageJson.version,
```
astandrik, line 146: Have you tried sth like window.ui_version? the same way it is used for react_app_disable_checks

Still feel that using package.json directly can have pitfalls.

config-overrides.js
```diff
@@ -48,13 +45,6 @@ module.exports = {
             }),
         );
 
-        // Add DefinePlugin to expose just the version
-        config.plugins.push(
-            new webpack.DefinePlugin({
-                'process.env.UI_VERSION': JSON.stringify(packageJson.version),
```
sareyu, line 54: what's wrong with this approach?

sareyu (commented):
LGTM

config-overrides.js
artemmufazalov: `process.env` values are embedded in build, so in build I have
`version: process.env.UI_VERSION` -> `version: 8.11.0`

However, when we use it as a package, we have only `process.env.UI_VERSION` in our code, when we import it and try to open in UI we receive `process` is not defined error, no variables are exposed

src/containers/UserSettings/settings.tsx
artemmufazalov: I tried `index.html`, the problem is that `webpack.DefinePlugin` does not expose vars to html files, so we can use them in js, but not in `index.html` (but all envs passed in command or taken from `.env` file work fine).

astandrik:
Can you try this approach?

https://github.com/ydb-platform/ydb-embedded-ui/pull/2005/files

artemmufazalov:
> Could you try this approach?
> 
> https://github.com/ydb-platform/ydb-embedded-ui/pull/2005/files

Please, describe, how it should work in case of package? There is no vars from `webpack.DefinePlugin` there

astandrik:
> > Could you try this approach?
> > https://github.com/ydb-platform/ydb-embedded-ui/pull/2005/files
> 
> Please, describe, how it should work in case of package? There is no vars from `webpack.DefinePlugin` there

You right - it wont work for package

Maybe 
```
import {version} from './package.json';
```
would work?

Here is some discussion on this topic https://stackoverflow.com/questions/9153571/is-there-a-way-to-get-the-version-from-the-package-json-file-in-node-js-code 

The idea of exposing entire package.json intuitively feels unsecure


However, our code is open-source and package.json is already visible in github to everyone - so its not a problem

Could we give it the last try with import {version} from './package.json'; and if it wont work - proceed with requiring the whole package.json

artemmufazalov:
> Could we give it the last try with import {version} from './package.json'; and if it wont work - proceed with requiring the whole package.json

There is another error
`Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)`

artemmufazalov:
> > Could we give it the last try with import {version} from './package.json'; and if it wont work - proceed with requiring the whole package.json
> 
> There is another error `Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)`

Actually, it works, but the error is displayed


Screenshot 2025-03-07 at 12 24 27

Screenshot 2025-03-07 at 12 23 42

astandrik (approved):
Well, at least we tried :_)

PR #1998, opened 2025-03-06
feat: create Api directly with YdbEmbeddedAPI class

artemmufazalov:


astandrik:
That only for good-looking purposes? Couldnt' really find reasonable matter

Just asking =D

sareyu (approved):

artemmufazalov:
> That only for good-looking purposes? Couldnt' really find reasonable matter
> 
> Just asking =D Maybe missing something

When I create Api in embedded ui, I use function `createApi` with some default values.  But if I want to extend it, there won't be defaults, and the params will be a little different

`createApi`
```js
// defaults
const api = createApi()
// with my values
const api = createApi({webVersion: false, withCredentials: false})
```
Current version with customization:
```js
class MyApi extends YdbEmbeddedAPI {
    customApi: MyCustomApi;

    constructor() {
        super({config: {withCredentials: true}, webVersion: true});

        this.customApi = new MyCustomApi();
    }
}

const api =  new MyApi()
```

So my goals:
- only one way to create Api without some proxy function with no clear purpose
- make params the same, so there will be similar approach both in embedded and in package
- make defaults the same

astandrik:
Thanks for explanation!

PR #2036, opened 2025-03-21
feat: parse logging link as default value

astandrik:


src/utils/logs.ts
```diff
@@ -0,0 +1,37 @@
...
+                return '';
+            }
+
+            const url = new URL(logUrl);
+
+            const queryParam = url.searchParams.get('query');
+            if (queryParam) {
+                const decodedQuery = decodeURIComponent(queryParam);
+
+                const updatedQuery = decodedQuery.replace(/\{([^}]*)\}/, (_match, contents) => {
```
artemmufazalov, line 24: Please, explain this with some example. Like before => after. In code comments, not here

src/utils/logs.ts
```diff
@@ -0,0 +1,37 @@
...
+                    const trimmedContents = contents.trim();
+                    return `{${trimmedContents}${trimmedContents ? ', ' : ''}database = "${dbName}"}`;
```
artemmufazalov, line 26:
```
if (trimmedContents){
  return`{${trimmedContents},  database="${dbName}"}`
} else {
  return `database="${dbName}"`
}
```
It looks better for me, ternaries inside template strings are hard to read

src/utils/__test__/logs.test.ts
```diff
@@ -0,0 +1,51 @@
...
+            url: 'https://monitoring.yandex-team.ru/projects/kikimr/logs?from=now-1h&to=now&query=%7B%7D',
+        };
+
+        const result = getLogsLink({
+            logging: JSON.stringify(loggingData),
+            dbName: 'testdb',
+        });
+
+        // Should add dbName to empty query
+        expect(result).toContain('database+%3D+%22testdb%22');
```
artemmufazalov, line 33: If query contains `database`, it doesn't mean, that it's correct. Please add also `expect` for the result for this test and for the first test.
```
expect(result).toBe("expected result string")
```

artemmufazalov (changes requested):

astandrik:
fixed

src/utils/logs.ts
astandrik: made code self-explanatory

src/utils/logs.ts
astandrik: made code self-explanatory

PR #2043, opened 2025-03-24
feat: show transfer errors in UI

nshestakov:
Info-—-Root-test-nginx_logs_trasfer-—-YDB-Monitoring-03-24-2025_05_01_PM

### DIFF
--- a/src/containers/Tenant/Diagnostics/Overview/TransferInfo/TransferInfo.tsx
+++ b/src/containers/Tenant/Diagnostics/Overview/TransferInfo/TransferInfo.tsx
@@ -5,6 +5,7 @@ import {AsyncReplicationState} from '../../../../../components/AsyncReplicationS
 import {YDBSyntaxHighlighter} from '../../../../../components/SyntaxHighlighter/YDBSyntaxHighlighter';
 import {YDBDefinitionList} from '../../../../../components/YDBDefinitionList/YDBDefinitionList';
 import {replicationApi} from '../../../../../store/reducers/replication';
+import type {DescribeReplicationResult} from '../../../../../types/api/replication';
 import type {TEvDescribeSchemeResult} from '../../../../../types/api/schema';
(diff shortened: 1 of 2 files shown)


src/store/reducers/replication.ts
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable camelcase */
```
artemmufazalov, line 1: It isn't needed in this file

src/containers/Tenant/Diagnostics/Overview/TransferInfo/TransferInfo.tsx
```diff
@@ -35,7 +38,29 @@ export function TransferInfo({data}: TransferProps) {
     );
 }
 
-function prepareTransferItems(data: TEvDescribeSchemeResult) {
+function prepareErrors(path: string, database: string) {
+    const {data, error} = replicationApi.useGetReplicationQuery({path, database}, {});
```
artemmufazalov, line 42: `replicationApi.useGetReplicationQuery` is a hook - it affects component rerenders. It is not an expected thing inside a prepare function. 

You should either move it to component level or make your functions hooks too. 

On Component level - most of our code is written this way
```js
function Component(){
  const {data: replicationData, error: replicationError} = replicationApi.useGetReplicationQuery({path, database}, {});
  const error = prepareErrors(replicationData, replicationError)
  const transferItems = prepareTransferItems(path, database, data, errors)
}
```

Or make everything hooks - rename both funcitons to `useSomething`
```js
function usePreparedErrors(path: string, database: string){
}
function usePreparedTransferItems((path: string, database: string, data: TEvDescribeSchemeResult){
  const errors = usePreparedErrors(path, database)
}
function Component(props){
  const transferItems = usePreparedTransferItems(path, database, data);
}


```

artemmufazalov (changes requested):

artemmufazalov (approved):

Issue #1948, opened 2025-02-13
Add delete DB button to Databases Table


artemmufazalov:
Figma: https://nda.ya.ru/t/mzzHWPCO7BwfE3

To Databases table add column with dropdown with ~~Connect, Edit and~~ Remove options

artemmufazalov:
"Connect to DB" looks not very clear in current dropdown - it's just snippet with connection string, while other options are DB can change DB, issue to improve idea and to add it to UI: https://github.com/ydb-platform/ydb-embedded-ui/issues/2099

Edit will be added in a separate issue: https://github.com/ydb-platform/ydb-embedded-ui/issues/2096

artemmufazalov:
Github PR: #2087 
Arcadia PR: https://nda.ya.ru/t/tEbtn1Tv7DKDj6

Issue #1940, opened 2025-02-12
Delete database in UI


artemmufazalov:
Figma: https://nda.ya.ru/t/drteadOn7BwfCJ and https://nda.ya.ru/t/ZJKyDD3j7BwfCs

What should be done:

- [x] Remove DB form with confirmation fields
- [x] Forms loading and error views, success and error toasters - reuse components from create DB form

~~- [ ] Manage button with dropdown (edit & remove) in header~~
~~- [ ] Edit form with the same resources as in create DB form~~

artemmufazalov:
Button to header on tenant page will be added here: https://github.com/ydb-platform/ydb-embedded-ui/issues/2100
Issue for edit DB: https://github.com/ydb-platform/ydb-embedded-ui/issues/2096

Issue #1939, opened 2025-02-12
Create database in UI


artemmufazalov:
Add create DB button and form to UI. It should use external meta handler.

Figma: https://nda.ya.ru/t/bokxo3ed7BweiQ

What should be done:
- [x] Create DB button in Databases Table
- [x] Form with DB setup
- [x] Form error view when DB cannot be created (lack of rights, not enough resources, etc.)
- [x] Create DB success toaster
- [x] Create DB error toaster

antonkovalenko:
Let's not forget to show:

- group kind - user must be able to select hdd/ssd for groups when backend allows
- number of groups to create. How many groups could be assigned to the database being created

artemmufazalov:
Swagger: 
https://nda.ya.ru/t/FupTw2IS7CuQwE

Protos: 
https://github.com/ydb-platform/ydb/blob/main/ydb/public/api/client/yc_private/ydb/v1/database_service.proto#L189
https://github.com/ydb-platform/ydb/blob/main/ydb/public/api/client/yc_private/ydb/v1/database.proto#L183

artemmufazalov:
Github PR: #2087 
Arcadia PR: https://nda.ya.ru/t/tEbtn1Tv7DKDj6

artemmufazalov:
Loading view will be added here: https://github.com/ydb-platform/ydb-embedded-ui/issues/2097

PR #2087, opened 2025-04-02
feat: allow pass create and delete DB funcs to UI

artemmufazalov:
How it will work when ui is used as a package: https://nda.ya.ru/t/a_NVBW7U7DHZKw
Stand with multi-cluster embedded version: https://nda.ya.ru/t/AUgj5s8F7DHZHW

Connected issues: #1939, #1940, #1948 

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/2087/)

  | 298 | 298 | 0 | 0 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: 🔺
  Current: 83.27 MB | Main: 83.25 MB
  Diff: +0.02 MB (0.02%)

  <details>

  </details>

### DIFF
--- a/src/components/TableWithControlsLayout/TableWithControlsLayout.scss
+++ b/src/components/TableWithControlsLayout/TableWithControlsLayout.scss
@@ -30,8 +30,6 @@
     &__table {
         position: relative;
         z-index: 2;
-
-        width: max-content;
     }
 
     .ydb-paginated-table__head {
--- a/src/components/TableWithControlsLayout/TableWithControlsLayout.scss
+++ b/src/components/TableWithControlsLayout/TableWithControlsLayout.scss
@@ -30,6 +30,8 @@
     &__table {
         position: relative;
         z-index: 2;
+
+        width: max-content;
     }
 
     .ydb-paginated-table__head {
--- a/src/containers/App/Content.tsx
+++ b/src/containers/App/Content.tsx
@@ -17,6 +17,8 @@ import {
     useCapabilitiesLoaded,
     useCapabilitiesQuery,
     useClusterWithoutAuthInUI,
+    useMetaCapabilitiesLoaded,
+    useMetaCapabilitiesQuery,
 } from '../../store/reducers/capabilities/hooks';
 import {nodesListApi} from '../../store/reducers/nodesList';
 import {cn} from '../../utils/cn';
@@ -213,8 +215,11 @@ function GetCapabilities({children}: {children: React.ReactNode}) {
     useCapabilitiesQuery();
     const capabilitiesLoaded = useCapabilitiesLoaded();
 
+    useMetaCapabilitiesQuery();
+    const metaCapabilitiesLoaded = useMetaCapabilitiesLoaded();
+
     return (
-        <LoaderWrapper loading={!capabilitiesLoaded} size="l">
+        <LoaderWrapper loading={!capabilitiesLoaded || !metaCapabilitiesLoaded} size="l">
             {children}
         </LoaderWrapper>
     );
--- a/src/containers/App/Content.tsx
+++ b/src/containers/App/Content.tsx
@@ -216,6 +216,7 @@ function GetCapabilities({children}: {children: React.ReactNode}) {
     const capabilitiesLoaded = useCapabilitiesLoaded();
 
     useMetaCapabilitiesQuery();
+    // It is always true if there is no meta, since request finishes with an error
     const metaCapabilitiesLoaded = useMetaCapabilitiesLoaded();
 
     return (
--- a/src/containers/Tenants/Tenants.tsx
+++ b/src/containers/Tenants/Tenants.tsx
@@ -1,8 +1,10 @@
 import React from 'react';
 
+import {CirclePlus, TrashBin} from '@gravity-ui/icons';
 import type {Column} from '@gravity-ui/react-data-table';
 import DataTable from '@gravity-ui/react-data-table';
-import {Button} from '@gravity-ui/uikit';
+import type {DropdownMenuItem} from '@gravity-ui/uikit';
+import {Button, DropdownMenu, Icon} from '@gravity-ui/uikit';
 
 import {EntitiesCount} from '../../components/EntitiesCount';
 import {ResponseError} from '../../components/Errors/ResponseError';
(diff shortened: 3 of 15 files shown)


src/uiFactory/uiFactory.ts
```diff
@@ -0,0 +1,13 @@
+import type {UIFactory} from './types';
+
+const uiFactoryBase: UIFactory = {};
+
+export function configureUIFactory(overrides: UIFactory) {
```
artemmufazalov, line 5: Approach inspired by ytsaurus-ui

https://github.com/ytsaurus/ytsaurus-ui/blob/main/packages/ui/src/ui/UIFactory/index.tsx#L476

I think it is easier and more straightforward to register functions here than to pass it from props.

Function for monitoring and logs links could be added here too

src/containers/Tenants/Tenants.tsx
```diff
@@ -218,12 +299,16 @@ export const Tenants = ({additionalTenantsProps}: TenantsProps) => {
...
     return (
-        <TableWithControlsLayout>
-            <TableWithControlsLayout.Controls>{renderControls()}</TableWithControlsLayout.Controls>
-            {error ? <ResponseError error={error} /> : null}
-            <TableWithControlsLayout.Table loading={loading}>
-                {currentData ? renderTable() : null}
-            </TableWithControlsLayout.Table>
-        </TableWithControlsLayout>
+        <div className={b('table-wrapper')}>
+            <TableWithControlsLayout>
```
astandrik, line 303: is it possible to pass className directly to TableWithControlsLayout as a prop?

src/components/TableWithControlsLayout/TableWithControlsLayout.scss
```diff
@@ -30,6 +30,8 @@
     &__table {
         position: relative;
         z-index: 2;
+
+        width: max-content;
```
astandrik, line 34: I am a liitle worried of this style possible impact on other tables

could you please clarify why this style was added

src/containers/App/Content.tsx
```diff
@@ -213,8 +215,11 @@ function GetCapabilities({children}: {children: React.ReactNode}) {
     useCapabilitiesQuery();
     const capabilitiesLoaded = useCapabilitiesLoaded();
 
+    useMetaCapabilitiesQuery();
+    const metaCapabilitiesLoaded = useMetaCapabilitiesLoaded();
```
astandrik, line 219: Did not dig deep but looks like if there is no meta metaCapabilitiesLoaded will always be falsy and loading will always be true?

src/store/reducers/capabilities/hooks.ts
```diff
@@ -87,3 +89,25 @@ export const useClusterWithoutAuthInUI = () => {
...
 };
+
+export function useMetaCapabilitiesQuery() {
+    capabilitiesApi.useGetMetaCapabilitiesQuery({});
+}
+
+export function useMetaCapabilitiesLoaded() {
+    const {data, error} = useTypedSelector(selectMetaCapabilities);
+
+    return Boolean(data || error);
```
astandrik, line 100: as for me its not that obvious that Loaded is true when we receive error

src/containers/Tenants/Tenants.tsx
artemmufazalov: TableWithControlLayout has `min-width: 100%`, parent container here have almost full screen width, I need to limit it to place button in the right corner of a table, not the right cornet of the screen

Without wrapper: 
Screenshot 2025-04-03 at 15 19 05

With wrapper:
Screenshot 2025-04-03 at 15 19 12

src/utils/hooks/useDatabaseFromQuery.ts
```diff
@@ -3,6 +3,13 @@ import {StringParam, useQueryParam} from 'use-query-params';
 export function useDatabaseFromQuery() {
     const [database] = useQueryParam('database', StringParam);
 
-    // Remove null from database type
+    // Remove null from type
     return database ?? undefined;
 }
+
+export function useClusterNameFromQuery() {
```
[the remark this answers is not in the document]
artemmufazalov: For me - yes, I prefer to import 1 hook instead of several imports from lib

src/store/reducers/capabilities/hooks.ts
artemmufazalov: For capabilities queries error is a valid state and it means that there is no capability endpoint in current cluster / meta cluster. In that case all capabilities considered not available (version 0)

The alternative for this code (return `data: {}` instead of `error` in catch block):
```js
 getMetaCapabilities: build.query({
            queryFn: async () => {
                try {
                    if (!window.api.meta) {
                        throw new Error('Method is not implemented.');
                    }
                    const data = await window.api.meta.getMetaCapabilities();
                    return {data};
                } catch {
                    // If capabilities endpoint is not available, there will be an error
                    // That means no new features are available
                    return {data: {}};
                }
            },
        }),
```

What do you think, will it be more clear?

astandrik: Do we want to handle real errors from this endpoint?

I mean now HTTP 500 and non-existent state are the same data: {}

artemmufazalov: I don't think so. 

In case of error we don't know anything about cluster capabilities, so it's much safer to consider everything disabled. 

If response is 500, we can only retry, but we do not retry requests currently

astandrik: as for me I would split loaded and availability state like

isMetaCapabilitiesAvailable=Boolean(window.api.meta.getMetaCapabilities)

loading={!isCapabilitiesLoaded || (isMetaCapabilitiesAvailable && !isMetaCapabilitiesLoaded)}

because when I see just !isMetaCapabilitiesLoaded I am expecting for it to be loaded


but thats just me and just readability issue

astandrik (dismissed):

src/store/reducers/capabilities/hooks.ts
artemmufazalov: I don't want to add additional code, but added some comments, I hope they will help

astandrik (approved):

PR #2125, opened 2025-04-09
feat(Tenants): correct links with relative path in balancer

### LINKED ISSUE #2105: [Build]: enable UI to work with relative balancer value
Current UI work only with absolute path in balancer. It should be capable to process values like `/proxy/host/ydb-vla-dev02-000.search.yandex.net` - using path relative to current host if there is no `REACT_META_BACKEND` or add path to meta cluster to balancer

artemmufazalov:
Closes #2105

Changes:
- Fix links to Tenant page from Databases table with relative balancer
- Fix links to Developer UI with relative balancer
- Do not create links with nodes hosts, only with balancer

#### How to check changes
Add following changes to `.env` file and run app with `npm run start`
```
##### EM #####
REACT_APP_BACKEND=
REACT_APP_META_BACKEND=http://ui-dev-0.ydb.yandex.net/api/meta8780
META_YDB_BACKEND=
```

`YDB DEV VLA02` and `klg1-0561` have paths relative to meta as balancer and these paths are also a little different. UI without this fix makes incorrect links to Tenant page and Developer UI.

`YDB RU PRESTABLE` has full balancer value, it works on current version and should work after fix.

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/2125/)

  | 312 | 308 | 0 | 4 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: ✅
  Current: 83.29 MB | Main: 83.29 MB
  Diff: +1.54 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/src/store/reducers/tenants/utils.ts
+++ b/src/store/reducers/tenants/utils.ts
@@ -15,15 +15,6 @@ const getControlPlaneValue = (tenant: TTenant) => {
     return controlPlaneName ?? defaultValue;
 };
 
-const getTenantBackend = (tenant: TTenant) => {
-    const node = tenant.Nodes ? tenant.Nodes[0] : {};
-    const address =
-        node.Host && node.Endpoints
-            ? node.Endpoints.find((endpoint) => endpoint.Name === 'http-mon')?.Address
-            : undefined;
-    return node.Host ? `${node.Host}${address ? address : ''}` : undefined;
-};
-
 export interface TenantMetricStats<T = string> {
     name: T;
     usage?: number;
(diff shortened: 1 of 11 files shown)


src/types/api/tenant.ts
```diff
@@ -45,7 +45,6 @@ export interface TTenant {
     StorageAllocatedSize?: string; // Actual database size
     /** uint64 */
     StorageMinAvailableSize?: string;
-    Nodes?: TSystemStateInfo[];
```
artemmufazalov, line 48: `Nodes` and `MonitoringEndpoint` requires additional param in endpoint (`nodes=true`), we don't sent this param and thus never have these fields in response

src/store/reducers/tenants/utils.ts
```diff
@@ -174,9 +165,8 @@ const calculateTenantEntities = (tenant: TTenant) => {
     return {nodesCount, groupsCount};
 };
 
-export const prepareTenants = (tenants: TTenant[], useNodeAsBackend: boolean): PreparedTenant[] => {
+export const prepareTenants = (tenants: TTenant[]): PreparedTenant[] => {
```
astandrik, line 168: As I see we used to use nodeId for backend - this was determined by singleClusterMode in tenants.ts

Why we are not using this approach now?

astandrik: Oh I see - we are not using tenant.backend field somewhere above...

astandrik (approved):
I'm sure we discussed this pr more than enough in pm =)

PR #2287, opened 2025-05-15
fix(Tenants): do not add empty search to query

artemmufazalov:
  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/2287/)

  | 318 | 316 | 0 | 2 | 0 |

  😟 No changes in tests. 😕

  ### Bundle Size: ✅
  Current: 83.53 MB | Main: 83.53 MB
  Diff: +0.12 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/src/store/state-url-mapping.ts
+++ b/src/store/state-url-mapping.ts
@@ -73,6 +73,7 @@ export const paramSetup = {
     '/cluster/tenants': {
         search: {
             stateKey: 'tenants.searchValue',
+            initialState: '',
         },
     },
 } as const;


astandrik:
image

I think we can do this with the same effect

astandrik (changes requested):
image

I think we can just do this with the same effect

artemmufazalov:
> image
> 
> I think we can just do this with the same effect

Empty string is the reason, why we have `search=` in query. If it is `undefined` there is no param, if it is empty string - there will be empty param

artemmufazalov:
The problem I solve is `search=` in query

astandrik:
image
I dont have search= in query with initialState: ''

https://github.com/user-attachments/assets/e4b84ec6-f141-45ae-9fad-8e146f480d9f

astandrik:
@artemmufazalov I understood your task and spent some time searching for an alternative more concise solution =)

I think setting initialState to empty string is this solution

as with initialState: '' the moment the value becomes an empty string the library thinks “we’re back to default” and removes search from location.search.

astandrik (approved):

PR #2335, opened 2025-05-27
fix: scrolling performance optimizations for table

astandrik:


Raubzeug (approved):

astandrik:
@Raubzeug actually think I'm done here
Stand redeployed

artemmufazalov (approved):

src/components/PaginatedTable/requestBatcher.ts
```diff
@@ -0,0 +1,182 @@
...
+    private async processBatch(): Promise<void> {
+        const allQueues = Array.from(this.requestQueue.entries());
+        this.requestQueue.clear();
+        this.batchTimeout = null;
+
+        for (const [_batchKey, requests] of allQueues) {
+            const groups = this.groupConsecutiveRequests(requests);
+
+            // Execute each group (consecutive chunks) as a separate batch
+            await Promise.all(groups.map((group) => this.executeBatch(group)));
```
astandrik, line 176: I suppose it should be allSettled here - if one group fails it doesnt mean everybody should fail

astandrik: there is try catch block inside executeBatch - so I suppose this case is already handled

PR #2452, opened 2025-06-20
fix(Cluster): use cluster name if title empty string

artemmufazalov:


src/store/reducers/cluster/cluster.ts
```diff
@@ -147,7 +147,7 @@ export function useClusterBaseInfo() {
     // Name: ydb_vla_dev02
     // Title: YDB DEV VLA02
     const clusterName = name ?? clusterNameFromQuery ?? undefined;
-    const clusterTitle = title ?? clusterName;
+    const clusterTitle = title || clusterName;
```
artemmufazalov, line 150: Cluster `title` may be an empty string.

`??` makes `clusterTitle` an empty string but `clusterName` should be used for correct display

Raubzeug (approved):

PR #2600, opened 2025-07-23
fix: use svg instead of mask for doughnuts

astandrik:


astandrik:
bugbot test

astandrik:
copilot hello

PR #2926, opened 2025-09-24
fix(HoverPopup): prevent untimely closing

### LINKED ISSUE #2787: Make vdisk/pdisk popup to be accessible for slower mouse movers
When mouse comes over vdisk or pdisk popup is displayed
It has links, users often move their mouse from vdisk/pdisk element to the respective popup.

According to feedback some users don't make it due to the blazing speed of popup disappearing.

https://nda.ya.ru/t/400S1UGu7JMHeX

Let's make UI more accessible.

Raubzeug:
closes #2787
[Stand](https://nda.ya.ru/t/hA-cREku7KMkzg)

### DIFF
--- a/src/components/VDisk/VDisk.tsx
+++ b/src/components/VDisk/VDisk.tsx
@@ -46,7 +46,7 @@ export const VDisk = ({
             onShowPopup={onShowPopup}
             onHidePopup={onHidePopup}
             renderPopupContent={() => <VDiskPopup data={data} />}
-            offset={{mainAxis: 5, crossAxis: 0}}
+            offset={{mainAxis: 2, crossAxis: 0}}
             delayClose={delayClose}
             delayOpen={delayOpen}
         >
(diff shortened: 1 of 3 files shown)


src/components/VDisk/VDisk.tsx
```diff
@@ -46,7 +46,7 @@ export const VDisk = ({
             onShowPopup={onShowPopup}
             onHidePopup={onHidePopup}
             renderPopupContent={() => <VDiskPopup data={data} />}
-            offset={{mainAxis: 5, crossAxis: 0}}
+            offset={{mainAxis: 2, crossAxis: 0}}
```
artemmufazalov, line 49: Please, explain, how your changes make everything better? I see no difference

Raubzeug: The main difference is that YDB_POPOVER_CLASS_NAME is now applied not to the popup itself, but to its content. Previously, when paddings were set at the popup level, the popup would close when the mouse hovered over these paddings. Now, paddings are part of the content, so the popup no longer closes prematurely.
I also made some minor adjustments to the offsets, but that's not the main improvement.

artemmufazalov (approved):

PR #3149, opened 2025-12-01
fix: format bytes columns and avoid refetch on column toggle

DaryaVorontsova:
Stand: https://nda.ya.ru/t/VsEVE-UU7Nw5yn

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/3149/)

  | 378 | 375 | 0 | 1 | 2 |

  
  <details>
  <summary>Test Changes Summary ⏭️2 </summary>

  #### ⏭️ Skipped Tests (2)
1. Scroll to row, get shareable link, navigate to URL and verify row is scrolled into view (tenant/diagnostics/tabs/queries.test.ts)
2. Copy result button copies to clipboard (tenant/queryEditor/queryEditor.test.ts)

  </details>

  ### Bundle Size: ✅
  Current: 62.33 MB | Main: 62.33 MB
  Diff: +2.88 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/.github/workflows/publication.yml
+++ b/.github/workflows/publication.yml
@@ -4,6 +4,11 @@ on:
 
 name: Publication
 
+permissions:
+  contents: write
+  id-token: write
+  packages: write
+
 jobs:
   publication:
     runs-on: ubuntu-latest
@@ -18,7 +23,8 @@ jobs:
       - run: npm ci
       - run: npm publish
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
+          NODE_AUTH_TOKEN: ''
+          NPM_TOKEN: ''
       - run: npm run build:embedded:archive
       - run: gh release upload ${{ github.event.release.tag_name }} $ASSET_NAME.zip
         env:
--- a/src/containers/Tenant/utils/schemaQueryTemplates.ts
+++ b/src/containers/Tenant/utils/schemaQueryTemplates.ts
@@ -24,10 +24,6 @@ function toLF(str: string) {
     return str.replace(/\r\n?/g, '\n');
 }
 
-function indentBlock(str: string, pad = '    ') {
-    return str.replace(/^/gm, pad);
-}
-
 export const createTableTemplate = (params?: SchemaQueryParams) => {
     const tableName = params?.relativePath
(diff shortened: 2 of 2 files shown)


src/components/PaginatedTable/TableChunksRenderer.tsx
```diff
@@ -125,6 +127,7 @@ export const TableChunksRenderer = <T, F>({
                     shouldFetch={chunkState.shouldFetch}
                     shouldRender={chunkState.shouldRender}
                     keepCache={keepCache}
+                    useColumnsIdsInRequest={useColumnsIdsInRequest}
```
astandrik: Currently we pass useColumnsIdsInRequest down the components tree to determine fetch options

The question is - do we use the same entities table with and without useColumnsIdsInRequest simultenously?

If not - I suggest to move this to fetching logics

DaryaVorontsova: If I understood you correctly, you suggest not to pass the useColumnsIdsInRequest prop down the component tree, but instead decide whether to include columnsIds in the fetch options inside the fetching logic itself.
In my case most tables always use column IDs in the request, and the peers table never uses them. So I can do something like:
`const shouldUseColumnsIds = tableName !== 'node-peers';` (or use a map of table names) directly in the fetching logic instead of passing a prop from the UI.

Is that what you meant?

astandrik:
btw is it intentional that in NODE_PEERS column order changing triggers refetch?

If its not - we could completely erase these checks and always preserve columns order for fetching

DaryaVorontsova:
Actually I had to avoid refetching every time when we add or remove a column in the table of peers, because backend for /peers doesn't support passing "required columns" (unlike other handlers like /nodes, etc.). And my idea is: if we have the peers table, we keep the columnsIds array unchanged (it is always empty) -> changing visible columns doesn’t trigger a new request to the backend )))

astandrik:
Okay now I see, thanks
In this case my comment about renaming is incorrect - hasColumnsIdsInRequest was better - because we always preserve order when we use columnIds

enums for table names is great idea though - much better than strings

Raubzeug (approved):

PR #3999, opened 2026-06-11
feat: support execute-query-and-forget for compaction

### LINKED ISSUE #3980: Support execute-query-and-forget action for compaction
Backend issue https://github.com/ydb-platform/ydb/pull/42592

Raubzeug:
[Stand](https://nda.ya.ru/t/0-yPVT0g7f3HGf)

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/3999/?t=1781871254350)

  | 686 | 679 | 0 | 4 | 3 |

  
  <details>
  <summary>Test Changes Summary 🗑️1</summary>

  #### 🗑️ Deleted Tests (1)
1. on: bridge piles visual states (bridge/bridge.test.ts)
  </details>

  ### Bundle Size: 🔺
  Current: 64.17 MB | Main: 64.16 MB
  Diff: +7.11 KB (0.01%)

  <details>

  </details>


Raubzeug:
@greptile-review

src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx
```diff
@@ -262,20 +286,63 @@ function CompactTableDialog({
...
+                showCompactionToast(
+                    result === START_COMPACTION_TIMEOUT_RESULT
+                        ? i18n('toast_compaction-started')
+                        : i18n('toast_compaction-completed'),
+                );
                 handleClose();
+
+                if (result === START_COMPACTION_TIMEOUT_RESULT) {
+                    onRefreshCompactions?.();
+                }
```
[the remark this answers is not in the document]
Raubzeug: I think this promise is still observed: [Promise.race()](vscode-webview://0mkvi68mueh9etj1e563qvbklkjclk4rn7h839q8jjfisnjj49j3/src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx:308) attaches handlers to the promise returned by [applyPromise.then()](vscode-webview://0mkvi68mueh9etj1e563qvbklkjclk4rn7h839q8jjfisnjj49j3/src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx:309), so a later rejection after the timeout should not surface as unhandled. The timeout branch intentionally means we stop waiting and treat compaction as started in background.

src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx
```diff
@@ -262,20 +284,60 @@ function CompactTableDialog({
...
                     parallel: parsedParallel,
                 });
+
+                if (executeQueryAndForgetAvailable) {
+                    // Backend returns immediately once compaction is scheduled in the background
+                    await applyPromise;
+                } else {
+                    // Compaction query is long-running, so do not block the dialog on it.
+                    // Treat reaching the response timeout without an error as a success.
+                    const result = await Promise.race([
```
astandrik: Is this fallback intentionally best-effort for old clusters? When `executeQueryAndForgetAvailable` is false, we still send a regular `execute-query`, but a 1s local timeout closes the dialog and shows the success toast before that request completes. If this tradeoff is intended for long-running compactions, maybe the UI copy should say the compaction request was sent/scheduled rather than confirming that compaction has started, or this late-error tradeoff should be documented here.

astandrik (dismissed):
Requesting changes for the old-cluster compaction fallback. The new timeout path can show success before the legacy `execute-query` request has actually completed, hiding late backend errors from the user.

astandrik (dismissed):

src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx
```diff
@@ -262,20 +284,60 @@ function CompactTableDialog({
...
+                        applyPromise
```
astandrik: I would also suggest to move these logics to some wrapper like

waitForCompactionStartOrTimeout - not to mix abstraction levels in this if-else case

like 

```
type CompactionStartResult =
    | {status: 'success'}
    | {status: 'error'; error: unknown}
    | {status: 'timeout'};

function waitForCompactionStartOrTimeout(
    promise: Promise<void>,
    timeoutMs: number,
): Promise<CompactionStartResult> {
    return new Promise((resolve) => {
        const timeoutId = window.setTimeout(() => {
            resolve({status: 'timeout'});
        }, timeoutMs);

        promise
            .then(() => {
                window.clearTimeout(timeoutId);
                resolve({status: 'success'});
            })
            .catch((error: unknown) => {
                window.clearTimeout(timeoutId);
                resolve({status: 'error', error});
            });
    });
}

// ...

if (executeQueryAndForgetAvailable) {
    // Backend returns immediately once compaction is scheduled in the background
    await applyPromise;
} else {
    // Compaction query is long-running, so do not block the dialog on it.
    // Treat reaching the response timeout without an error as a success.
    const result = await waitForCompactionStartOrTimeout(
        applyPromise,
        START_COMPACTION_RESPONSE_TIMEOUT,
    );

    if (result.status === 'error') {
        setRequestErrorMessage(prepareErrorMessage(result.error));
        return;
    }
}


......

```

src/containers/Tenant/Diagnostics/Overview/TableInfo/CompactTableAction/CompactTableAction.tsx
Raubzeug: The fallback is intentional and the "compaction started" copy is correct for old clusters. The semantics there are: the request is sent and the operation starts in the background; if the operation fails to start, the request errors out — so if the request doesn't error, the operation has started. That's why confirming "compaction started" is accurate rather than just "scheduled/sent".

The 1s timeout is only there so the dialog doesn't block on the long-running query: within that window we still surface a real error and keep the dialog open; if no error arrives, we treat it as started and close. The one edge case is an error arriving after the 1s timeout — that won't be shown since the dialog is already closed, but user will see it in label with compaction operation status.

astandrik (approved):

PR #4026, opened 2026-06-18
feat(TopicData): support new meta handler for data

### LINKED ISSUE #3990: Make UI send request to new YDB EM schema_topic_data handler instead of topic_data YDB viewer in Data page
Tracker ticket: LOGBROKER-10467

Raubzeug:
[Stand](https://nda.ya.ru/t/RfFWD1pW7fjNuf)
closes #3990

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/4026/?t=1781866283995)

  | 688 | 681 | 0 | 4 | 3 |

  😟 No changes in tests. 😕

  ### Bundle Size: 🔺
  Current: 64.16 MB | Main: 64.14 MB
  Diff: +0.02 MB (0.03%)

  <details>

  </details>


Raubzeug:
@greptile-review

src/containers/Tenant/Diagnostics/TopicData/TopicData.tsx
```diff
@@ -166,11 +170,21 @@ export function TopicData({scrollContainerRef, path, database, databaseFullPath}
         () => ({
             path,
             database,
+            clusterName,
             partition: selectedPartition ?? '',
             isEmpty: emptyData,
             currentPage,
+            useMeta: schemaTopicDataAvailable,
```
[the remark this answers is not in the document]
Raubzeug: no need to use new handler for probe

src/containers/Tenant/Diagnostics/TopicData/columns/columns.tsx
```diff
@@ -323,38 +327,89 @@ export function TopicDataTsDiff({
...
     let invalid = false;
-    try {
-        encryptedMessage = atob(message);
-    } catch {
-        encryptedMessage = i18n('description_failed-decode');
-        invalid = true;
+    if (typeof message === 'string') {
+        messageToRender = message;
+        try {
+            messageToRender = atob(message);
```
astandrik: This still treats every string as legacy viewer data and tries to base64-decode it. That is correct for `/viewer/topic_data`, but the new meta path can return `Message` as an already schematized JSON value, including string primitives.

Repro:
1. Enable Topic Data through `/meta/schema_topic_data`.
2. Return a schematized message with a string primitive, for example `Message: "hello"`.
3. Open the Topic Data table and then the drawer for that row.

Expected: the UI renders `hello` as the message value.
Actual: the table tries `atob("hello")` and shows a decode error or corrupted text. The drawer has the same problem in `TopicMessage`.

Please keep an explicit signal for the response source/schematization state, such as `useMeta` or `isSchematized`, and skip base64 decoding for schematized values, including string primitives.

src/containers/Tenant/Diagnostics/TopicData/TopicData.tsx
astandrik, line 177: The table fetch now uses the meta endpoint, but the probe before it still does not. `useTopicDataAvailable()` can enable this screen from `/meta/schema_topic_data`, and `tableFilters`/drawer pass `useMeta` and `clusterName`, but `useTopicProbeQuery()` is still called without those values and therefore falls back to the legacy viewer request.

Repro:
1. Use a cluster where `/meta/schema_topic_data` is available but `/viewer/topic_data` is unavailable or blocked.
2. Open Topic Data.
3. Select a partition and apply a timestamp filter.
4. Watch the initial probe request.

Expected: the probe uses the same meta path as the table fetch.
Actual: the probe calls the legacy viewer path, so timestamp-based offset resolution can fail even though the table fetch itself supports meta.

Please pass the same `useMeta` and `clusterName` into `useTopicProbeQuery()`, or do not advertise Topic Data availability from `/meta/schema_topic_data` until every fetch path uses it.

astandrik (changes requested):
Requesting changes for two Topic Data regressions. The table/drawer still cannot distinguish schematized string values from legacy base64 strings, and the timestamp/offset probe still uses the legacy viewer path even when Topic Data is enabled through the meta capability.

src/containers/Tenant/Diagnostics/TopicData/TopicData.tsx
Raubzeug: I deliberately not using /meta/schema_topic_data for probe query, cause /meta/schema_topic_data is little bit slower than viewer/topic_data.

src/containers/Tenant/Diagnostics/TopicData/columns/columns.tsx
Raubzeug: The tricky part is that the new /meta/schema_topic_data endpoint can return either a schematized value or a non-schematized (legacy-style) one, so a single useMeta/isSchematized flag per response isn't enough — we can't know upfront whether a given Message will be a parsed object/primitive or a base64-encoded string.

src/containers/Tenant/Diagnostics/TopicData/TopicData.tsx
astandrik: I understand the performance reason, and using `/viewer/topic_data` for the probe is fine when that endpoint is actually available.

The regression is about the meta-only case. `useTopicDataAvailable()` now enables Topic Data when `/meta/schema_topic_data` is available even if `/viewer/topic_data` is not. In that configuration the table fetch can use meta, but the probe still goes to viewer and may fail before timestamp-based offset resolution works.

So we need one of these:

1. use viewer for the probe only when `/viewer/topic_data` is available, and fall back to `/meta/schema_topic_data` for meta-only clusters;
2. or do not enable Topic Data from `/meta/schema_topic_data` alone.

The slower meta probe is still better than a broken probe on clusters where viewer topic data is unavailable.

src/containers/Tenant/Diagnostics/TopicData/columns/columns.tsx
astandrik: That makes sense, and I agree that a single response-level `useMeta` flag is not enough if `/meta/schema_topic_data` can return mixed message formats.

But that is exactly the issue here: a string value is ambiguous. From the frontend side, both cases have the same shape:

- schematized string primitive: `Message: "hello"`
- legacy-style encoded payload: `Message: "<base64 string>"`

So `typeof message === "string"` cannot tell us which path to use. Decoding every string preserves legacy messages but breaks valid schematized string primitives. Treating every string as plain text would break legacy messages.

We need a per-message signal from the response contract. For example:

- `MessageEncoding: "base64" | "json"` / `MessageFormat`
- `IsSchematized: boolean`
- separate fields, e.g. `MessageBase64` for legacy payload and `Message` for schematized JSON
- or, if this is already guaranteed, use `SchematizeError` as the signal: decode only when the message failed schematization and stayed legacy-style.

Without such a per-message signal this cannot be fixed reliably in the UI, because some plain strings can also be valid base64-looking strings.

astandrik (changes requested):

src/containers/Tenant/Diagnostics/TopicData/TopicData.tsx
Raubzeug: It's impossible case when /meta/schema_topic_data is available, but /viewer/topic_data is not. /meta/schema_topic_data handler gets data from /viewer/topic_data directly and after that fetches additional scheme info.

astandrik: I rechecked the ydb-em handler, and you are right about this path.

`/meta/schema_topic_data` first proxies the request to `/viewer/topic_data` and only then adds schema-registry decoding. If that viewer request fails, the meta handler fails as well. So the case I described, where the table fetch works through meta while the probe fails only because it uses viewer, is not accurate.

Keeping the probe on `/viewer/topic_data` for performance is fine under this contract.

src/containers/Tenant/Diagnostics/TopicData/columns/columns.tsx
Raubzeug: Great catch, now I see the issue!

astandrik (approved):

PR #4301, opened 2026-09-02
fix: if no cluster name redirect to root

### LINKED ISSUE #2352: fix: e.reduce is shown when user uses invalid url for clusters
Now user sees 

```
TypeError: e.reduce is not a function
    at f (/monitoring/static/js/main.7e4cf3c9.js:2:1100909)
```

However in most cases logics of this error is awaited by us (user didnt passed a cluster or used some invalid url) - we could show more verbose and comprehensible error

Raubzeug:
closes #2352

<h3>Greptile Summary</h3>

This change adds a guard for multi-cluster cluster URLs that lack a usable cluster name and adds focused coverage for the guard predicate.

The reported redirect concern was disproved through a rendered `Content` route test starting at `/cluster?clusterName=` in multi-cluster mode. The application ultimately navigated to `/home/:activeTab` with the query string cleared, and the header, main content area, and home page rendered normally. The intermediate `/` redirect is handled by the existing fallback route, so it does not produce an empty page.

<h3>Confidence Score: 5/5</h3>

Safe to merge based on the verified routing behavior; no actionable findings remain.

There are no final scoring findings. The reported failure path was exercised in a rendered application harness and produced the expected home-page content rather than an empty view.

**Files Needing Attention:** None.

<sub>Reviews (1): Last reviewed commit: ["fix: if no cluster name redirect to root"](https://github.com/ydb-platform/ydb-embedded-ui/commit/e37328390881a908e8b95ceeeb159484bbbaa7fe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59518210)</sub>

  📊 [Full Report](https://ydb-platform.github.io/ydb-embedded-ui/4301/?t=1788420768335)

  | 950 | 948 | 0 | 2 | 0 |

  
  <details>
  <summary>Test Changes Summary ✨1 🗑️9</summary>

  #### ✨ New Tests (1)
1. on: bridge piles visual states (bridge/bridge.test.ts)

#### 🗑️ Deleted Tests (9)
1. on: shows unavailable healthcheck as non-interactive unknown (bridge/bridge.test.ts)
2. on: uses scoped issues without assuming global pile support (bridge/bridge.test.ts)
3. on: opens the source healthcheck issue from a pile badge (bridge/bridge.test.ts)
4. on: scrolls to a deep-linked issue after delayed healthcheck loading (bridge/bridge.test.ts)
5. on: updates the selected tab when the target changes in an open drawer (bridge/bridge.test.ts)
6. on: shows combined bridge pile and healthcheck states (bridge/bridge.test.ts)
7. on: shows bridge pile healthcheck loading state (bridge/bridge.test.ts)
8. filters tablets by multiple types and tablet ID (cluster/tablets.test.ts)
9. keeps the type filter local during database-wide tablet ID search (cluster/tablets.test.ts)
  </details>

  ### Bundle Size: ✅
  Current: 65.61 MB | Main: 65.61 MB
  Diff: +1.78 KB (0.00%)

  ✅ Bundle size unchanged.

  <details>

  </details>

### DIFF
--- a/src/containers/App/Content.tsx
+++ b/src/containers/App/Content.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import {connect} from 'react-redux';
 import type {RedirectProps} from 'react-router-dom';
-import {Redirect, Route, Switch} from 'react-router-dom';
+import {Redirect, Route, Switch, useLocation} from 'react-router-dom';
 
 import {Unauthenticated} from '../../components/Errors/401';
 import {AccessDenied} from '../../components/Errors/403';
@@ -41,6 +41,7 @@ import {
     TenantSlot,
     VDiskPageSlot,
 } from './appSlots';
+import {shouldRedirectClusterRouteToRoot} from './clusterRouteGuard';
 import {getLegacyClusterTenantsRedirect, getLegacyTenantRedirect} from './legacyRedirects';
 
 import './App.scss';
@@ -148,6 +149,12 @@ export function Content(props: ContentProps) {
     const {singleClusterMode} = props;
 
     const slots = useSlots(props);
+    const currentLocation = useLocation();
+    const shouldRedirectToRoot = shouldRedirectClusterRouteToRoot({
+        singleClusterMode,
+        pathname: currentLocation.pathname,
+        search: currentLocation.search,
+    });
 
     const additionalRoutes = slots.get(RoutesSlot);
 
@@ -158,6 +165,7 @@ export function Content(props: ContentProps) {
     return (
         <Switch>
             {additionalRoutes?.rendered}
+            {shouldRedirectToRoot ? <Redirect to="/" /> : null}
             <Route>
                 <Header />
                 <Switch>
--- a/src/containers/App/clusterRouteGuard.ts
+++ b/src/containers/App/clusterRouteGuard.ts
@@ -1,0 +1,17 @@
+import {checkIsClusterPage} from '../../routes';
+
+export function shouldRedirectClusterRouteToRoot({
+    singleClusterMode,
+    pathname,
+    search,
+}: {
+    singleClusterMode: boolean;
+    pathname: string;
+    search: string;
+}) {
+    if (singleClusterMode || !checkIsClusterPage(pathname)) {
+        return false;
+    }
+
+    return !new URLSearchParams(search).get('clusterName');
+}
--- a/src/containers/App/clusterRouteGuard.ts
+++ b/src/containers/App/clusterRouteGuard.ts
@@ -1,4 +1,6 @@
-import {checkIsClusterPage} from '../../routes';
+import {matchPath} from 'react-router-dom';
+
+import routes from '../../routes';
 
 export function shouldRedirectClusterRouteToRoot({
     singleClusterMode,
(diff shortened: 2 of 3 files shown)


src/containers/App/Content.tsx
```diff
@@ -158,6 +165,7 @@ export function Content(props: ContentProps) {
     return (
         <Switch>
             {additionalRoutes?.rendered}
+            {shouldRedirectToRoot ? <Redirect to="/" /> : null}
             <Route>
```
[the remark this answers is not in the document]
Raubzeug: This is intentional. This redirect handles an invalid/incomplete cluster context rather than migrating the same logical page, so it redirects to the agreed root URL without explicitly carrying cluster-scoped query/hash state such as backend.
The legacy redirects preserve search and hash because they only rename a route while keeping the same resource context.
additionalRoutes also intentionally stays first: RoutesSlot is a public extension/override point, so consumer-provided routes should take precedence. If no slot handles the location, the guard still runs before the built-in cluster route and its data requests.

src/containers/App/clusterRouteGuard.ts
```diff
@@ -0,0 +1,17 @@
...
+export function shouldRedirectClusterRouteToRoot({
+    singleClusterMode,
+    pathname,
+    search,
+}: {
+    singleClusterMode: boolean;
+    pathname: string;
+    search: string;
+}) {
+    if (singleClusterMode || !checkIsClusterPage(pathname)) {
```
astandrik: The guard is stricter than the route it protects, so the original invalid-URL crash is still reachable.

Reproduction:
1. Run the app in multi-cluster mode.
2. Open `/cluster/databases/extra?databasePage=query` without `clusterName`.
3. The guard returns `false`, the non-exact cluster route mounts, and the error boundary shows `TypeError: nodesList?.reduce is not a function`.

References:
- [`checkIsClusterPage` uses `path-to-regexp` end matching](https://github.com/ydb-platform/ydb-embedded-ui/blob/e37328390881a908e8b95ceeeb159484bbbaa7fe/src/routes.ts#L284-L286).
- [The built-in cluster route has no `exact` flag](https://github.com/ydb-platform/ydb-embedded-ui/blob/e37328390881a908e8b95ceeeb159484bbbaa7fe/src/containers/App/Content.tsx#L58-L64).
- [Issue #2352 describes the original invalid-URL crash](https://github.com/ydb-platform/ydb-embedded-ui/issues/2352).

Please match the guard with the route's non-exact semantics and add a rendered regression for the trailing-segment case.

astandrik (changes requested):
Blocking: the guard does not cover every path accepted by the cluster route, so the failure reported in #2352 remains reachable.

src/containers/App/clusterRouteGuard.ts
astandrik: Verified on [a75a3223](https://github.com/ydb-platform/ydb-embedded-ui/commit/a75a32238f736586f172c210f36a0200166b0cc8). The trailing-segment URL now redirects to home without mounting cluster data loading. The rendered regression fails with the previous guard and passes with this fix. Guard tests (7/7) and the full suite (1,834 tests) pass. Resolving this thread.

astandrik (approved):

